PR #244, opened 2023-03-22
Cleaning up Console Warnings

### LINKED ISSUE #241: Console warnings
Our extension prints several serious looking but harmless warning messages. We should fix these.

image

weegeekps:
Resolves #241.

Trying to clean up some console warnings the best we can.


weegeekps:
@lilleyse I went ahead and opened this as a draft as I'm not ready to say we're done with this. You may also want to take a look since you're a bit more familiar with the Fabric code.

lilleyse:
Whoops, that bug was my mistake: https://github.com/CesiumGS/cesium-omniverse/commit/489f1ca56dd03018b7fcf6f69b2ec4fd9050db39. Thanks for fixing it.

PR #246, opened 2023-03-22
Fixed a crash when removing the last available token for a tileset

### LINKED ISSUE #226: Clearing token in the Cesium prim causes Create to hang
Clearing the Project Token field in the Cesium prim causes create to hang.

Note: This only happens if there is a tileset in the stage.

weegeekps:
Resolves #226.

We are crashing when the user removes the last available token if a tileset and imagery are loaded. This is due to us having some unchecked access to an optional value in `OmniTileset::addImageryIon`. I added a check to bail out early from this function since without a token there is no reason to do any of the other work that frame.

### DIFF
--- a/src/core/src/OmniTileset.cpp
+++ b/src/core/src/OmniTileset.cpp
@@ -277,6 +277,12 @@ void OmniTileset::addImageryIon(const pxr::SdfPath& imageryPath) {
     const OmniImagery imagery(imageryPath);
     const auto imageryIonAssetId = imagery.getIonAssetId();
     const auto imageryIonAccessToken = imagery.getIonAccessToken();
+
+    if (!imageryIonAccessToken.has_value()) {
+        // If we don't have an access token available there's no point in adding the imagery.
+        return;
+    }
+
     const auto imageryName = imagery.getName();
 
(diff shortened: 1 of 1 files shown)


src/core/src/OmniTileset.cpp
```diff
@@ -277,6 +277,12 @@ void OmniTileset::addImageryIon(const pxr::SdfPath& imageryPath) {
     const OmniImagery imagery(imageryPath);
     const auto imageryIonAssetId = imagery.getIonAssetId();
     const auto imageryIonAccessToken = imagery.getIonAccessToken();
+
+    if (!imageryIonAccessToken.has_value()) {
+        // If we don't have an access token available there's no point in adding the imagery.
+        return;
```
weegeekps, line 283: One open question I have is whether or not we should reprompt the user to add a default project token if no default token is available. @r-veenstra, not sure if you have any thoughts here regarding the user perspective?

lilleyse: I think this would be a case where we'd want to open the token troubleshooter, though I haven't checked what Unreal and Unity do. That could be a follow up PR.

PR #247, opened 2023-03-27
Implemented the search field in the asset window

### LINKED ISSUE #245: Implement search bar in Cesium Assets panel
Add a search bar to the Cesium Assets panel to filter results, similar to Unity and Unreal.

image

weegeekps:
Resolves #245.

Implemented the search field. Wasn't happy with the styling I originally came up with for it but after some tweaking I think I've settled on something that should be good enough for most users. Does a pretty basic substring filter looking for the exact sequence of characters in the name field. This is similar to what we do for Unity and Unreal, but we may want to consider doing a more powerful indexed search in the future.

https://user-images.githubusercontent.com/284269/228000428-5c36f351-e16d-4594-b291-d27c66f812fc.mp4

### DIFF
--- a/exts/cesium.omniverse/cesium/omniverse/ui/__init__.py
+++ b/exts/cesium.omniverse/cesium/omniverse/ui/__init__.py
@@ -11,3 +11,4 @@ from .fabric_modal import CesiumFabricModal  # noqa: F401
 from .credits_window import CesiumOmniverseCreditsWindow  # noqa: F401
 from .credits_viewport_frame import CesiumCreditsViewportFrame  # noqa: F401
 from .credits_parser import CesiumCreditsParser  # noqa: F401
+from .search_field_widget import CesiumSearchFieldWidget  # noqa: F401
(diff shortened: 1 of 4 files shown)


lilleyse:
The code looks good. I have a couple small nitpicks that don't need to hold up the PR.

* The search box seems a bit too big and the text isn't vertically centered
* The search bar vertical size increases slightly when you enter text (would it help if there were an .svg close icon?)
* The asset menu table sometimes flickers when a new filter is set

search

weegeekps:
> The code looks good. I have a couple small nitpicks that don't need to hold up the PR.

Don't approve it yet then, I want to see what I can do to fix some of these. Unfortunately the UI system leaves a lot to be desired for centering so I don't think I can fix that, but I do notice the vertical size increasing which I missed before.

lilleyse:
The close icon seems to be missing now.

image

lilleyse:
Also it might be a bit late to suggest this, but did you see if it was possible to use the search widget in the content panel? I wasn't sure if they made that widget available for use outside that panel.

image

weegeekps:
> Also it might be a bit late to suggest this, but did you see if it was possible to use the search widget in the content panel? I wasn't sure if they made that widget available for use outside that panel.

I looked at that when I was starting out. It's not possible.

weegeekps:
@lilleyse, this is ready for a final review pass.

PR #248, opened 2023-03-27
Added word_wrap to the Asset Details header

### LINKED ISSUE #223: Long tileset names hide the Add to Stage button in the Add UI
When selecting a dataset with a long name, the title makes description panel extra wide and often the Add to Stage button is hidden as it is pushed far off to the right of the screen.  The user can scroll horizontally to find it, however it is not intuitive.

image

weegeekps:
Resolves #223.

We were missing `word_wrap` on the label for the Asset Details panel header. This was causing long asset names to expand beyond the max size of the window.


r-veenstra:
Tested and looks good to me

PR #251, opened 2023-03-28
Added a placeholder into the create field of the token window

### LINKED ISSUE #164: Create a new token should have a placeholder token name
The create a new token textbox should have a placeholder such as: **cesium-for-omniverse-test (Created by Cesium For Omniverse)**

weegeekps:
Resolves #164.

Adds a default placeholder value into the create field of the token window. If the stage has been saved, this will be the filename of the USD. If the stage hasn't been saved, this will be the `anon:` string set automatically when the stage is created.

@r-veenstra, I'm wondering if you have any opinion on including the `anon:` value? My initial reaction was to just not set the placeholder in that case, but I relented and decided to include it.

#### Stage Unsaved
image

#### Stage Saved
image

### DIFF
--- a/exts/cesium.omniverse/cesium/omniverse/ui/token_window.py
+++ b/exts/cesium.omniverse/cesium/omniverse/ui/token_window.py
@@ -3,6 +3,7 @@ import carb.events
 import omni.kit.app as app
 import omni.ui as ui
 import omni.usd
+from pathlib import Path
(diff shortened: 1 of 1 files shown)


lilleyse:
Discussion about the `anon:` prefix aside - this works well for me

lilleyse (approved):

Issue #116, opened 2023-01-11
Replace `InMemoryAssetResolver` with `omni.ui.DynamicTextureProvider`


lilleyse:
Currently we use a custom AssetPackageResolver USD plugin for in-memory textures, but the Nvidia folks pointed us to `omni.ui.DynamicTextureProvider` which is new in Kit 104.1 and can do the same thing.

Example: [dynamic-texture-provider.zip](https://github.com/CesiumGS/cesium-omniverse/files/10395737/dynamic-texture-provider.zip)

The C++ header is `#include <omni/ui/ImageProvider/DynamicTextureProvider.h>` but it's referencing some private headers at the moment.

Issue #95, opened 2022-12-19
InMemoryAssetResolver plugin not being called


weegeekps:
A regression occurred with Kit 104.1 where the first texture in memory is being applied to all tiles. This appears to be due to the `InMemoryAssetResolver` plugin not being properly called from the Nvidia code. The Nvidia folks are investigating. This issue is just to track the issue as updates occur.

![image (1)](https://user-images.githubusercontent.com/284269/208474721-c9799b52-a7a9-4209-869a-390ea455185d.png)

PR #252, opened 2023-03-29
Replace InMemoryAssetResolver with DynamicTextureProvider

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/116

Removes `InMemoryAssetResolver` and uses Omniverse's built-in `DynamicTextureProvider` instead. This has several benefits:

* No longer need to convert raw pixels to BMP
* Textures now work when Cesium for Omniverse is first installed, without a restart [^1]
* Compressed textures and mipmaps are supported in `DynamicTextureProvider`. That can be a follow up PR.
* Pixel data can be swapped out dynamically. This means we don't need to create a new material when higher detail imagery comes in. Going a step further, this allows us to do basic material pooling which could significantly improve performance. I haven't actually tested this yet, it will be a follow up PR.
* I no longer see cyan flashes (texture loading) for 3D Tiles input. Red flashes still happen, but that's expected.
* Removed `/rtx-transient/resourcemanager/enableTextureStreaming = false` workaround from https://github.com/CesiumGS/cesium-omniverse/issues/95

This doesn't fix the crashes / hangs we've noticed. At least not all of them. I can still trigger hangs by reloading agi.usda repeatedly.

One potential blocker is that each time a texture is loaded a warning gets printed. I reached out to the Nvidia team about this.

```
2023-03-29 19:50:32  [Warning] [omni.hydra] Material parameter 'tex' was assigned an incompatible texture: 'dynamic://imagery_ion_2__11718102_184221417_4827478_456288989__11717949_310164846_4827677_90915967'
```

image

To do:

* [x] Fix Windows CI

[^1]: Though a restart is still required to register Cesium USD schemas


src/core/src/DynamicTextureProviderCache.cpp
```diff
@@ -0,0 +1,50 @@
...
+    bool removeItem = false;
+
+    {
+        // Make sure the accessor goes out of scope and the item is unlocked before removing the item
+        DynamicTextureProviderCache::Accessor accessor;
+        auto found = _items.find(accessor, name);
+        if (found) {
+            auto& referenceCount = accessor->second->_referenceCount;
+            referenceCount--;
+            if (referenceCount == 0) {
```
weegeekps, line 29: As long as `found` is held it blocks other threads, correct? I'm concerned about if there's a case where we could run into a situation where reference count could go negative.

weegeekps (commented):
Looks good. One open question.

src/core/src/DynamicTextureProviderCache.cpp
lilleyse: I believe `accessor` acts as a lock guard, so we shouldn't see any race conditions

weegeekps: Sounds good. This is good to go once you fix the Windows builds. I'll give it a quick once over when that's done.

weegeekps (approved):
:rocket:

lilleyse:
This PR doesn't seem to fix the blue dots artifact that shows up in Create when a material is loaded. Interestingly this doesn't happen in `cesium.omniverse.app.kit`.

blue-flickers

This is different from the cyan placeholder when loading textures, which is fixed in this PR.

weegeekps (approved):
Good to go.

PR #253, opened 2023-03-30
Use same kit version as create 2022.3.3

lilleyse:
Our build uses a slightly ahead version of Kit SDK. I changed it back to `275.6d591d71` which is what Create 2022.3.3 uses. Generally we should be in lockstep with the official releases, I must have accidentally overlooked this.

image


weegeekps (approved):
As long as this passes, feel free to merge.

Issue #237, opened 2023-03-21
Build release zip on CI instead of locally


lilleyse:
We should be building the release zip on CI instead of locally. This includes:

* Add packaging step to our GitHub Actions workflow that builds the `package` target and publishes the `.zip` to S3
* Add new GitHub Action configuration for Docker CentOS builds
* Update the [release guide](https://github.com/CesiumGS/cesium-omniverse/blob/main/docs/release-guide/README.md) with instructions to download packages from S3 instead of building locally

PR #254, opened 2023-03-31
Build release zip on CI instead of locally

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/237

This can be reviewed commit by commit:

* Add packaging step to our GitHub Actions workflow that builds the `package` target and publishes the `.zip` to S3
* Add new GitHub Action configuration for Docker CentOS builds
* Update the [release guide](https://github.com/CesiumGS/cesium-omniverse/blob/main/docs/release-guide/README.md) with instructions to download packages from S3 instead of building locally


weegeekps:
I have quite a few questions, including some initial questions:

- Do we always publish to S3 on every build?
  - If yes, do we actually need to?
  - If yes, does the current versioned build on S3 get overwritten with every build?
- Can we set this up to only execute on a release tag?

lilleyse:
@weegeekps yeah that's a good idea

weegeekps (approved):
LGTM, :shipit:

lilleyse:
For some reason the 0.4.0 build wasn't able to pick up the git tag. It was working locally... I'll need to investigate why this didn't happen on CI.

For now I manually edited the name of the zip file from `CesiumForOmniverse-Windows-8832399.zip` to `CesiumForOmniverse-Linux-v0.4.0.zip`

PR #257, opened 2023-04-13
Python Unit Testing

weegeekps:
This enables unit testing in the Python portions of the extension. The tests do not run in CI yet, as there are some open questions about whether or not RTX is required. It's also important to think of these less as unit tests and more as E2E tests, since they currently manipulate the UI directly. Further refactoring is required to lessen this dependency to give us a good set of abstract code to more easily test.

Two scripts have been created to run these:
- `scripts/run_python_unit_tests.{sh,bat}` - Runs the unit tests.
- `scripts/interactive_python_unit_tests.{sh,bat}` - Opens the unit tests in an interactive window. Useful for developing unit tests.

Also included in this PR are a few sample tests.


lilleyse:
I tried out both scripts. It looks good and seems to be consistent with what other extensions are doing for testing.

PR #260, opened 2023-04-20
Adding first UI tests

weegeekps:
This PR adds unit tests for the following:

* `AssetToAdd` model
* `ImageryToAdd` model
* `CesiumPassFailWidget`

The last of these is our first test using `OmniUiTest` and can be used as a template for how to do tests for other UI components. `OmniUiTest` uses snapshot images to help validate that everything is working visually, so we have to provide reference images as well. These can be generated by first running the test, letting it fail, and then copying that image over to the proper path.


lilleyse:
Do you know why CesiumPassFailWidget's `__destroy__` isn't getting covered?

image

lilleyse:
Other than that comment, looks great

weegeekps:
So. Two things: `__destroy__` is not a thing, oops, but `__del__` (the actual destructor) isn't good for this either, as Python is garbage collected and the destructor isn't guaranteed to run. So I overrode `Widget`'s `destroy` method as is correct and then added a destructor that calls that in the event that a caller doesn't properly clean up after themselves. I don't think that's an issue, so the destructor may be overkill here, but it won't hurt either.

lilleyse:
Yup that fixed it. Thanks!

PR #262, opened 2023-04-26
Statistics widget

lilleyse:
Adds a statistics widget to the Cesium Debugging window that shows number of materials and geometries in the scene. 

Originally I was going to make this a viewport overlay but I think it's fine in the Cesium Debugging window where the UI style isn't so important and there's more space. Also, there's a cost to gathering these statistics each frame. It only runs if the Cesium Debugging window is in the dock.

Aside from that I also added some stuff to help with Python debugging. I wanted to enable the omni debug extensions by default in our test app but it doesn't dock by default. So I left those extensions commented out.

statistics

### DIFF
--- a/exts/cesium.omniverse/cesium/omniverse/ui/statistics_widget.py
+++ b/exts/cesium.omniverse/cesium/omniverse/ui/statistics_widget.py
@@ -1,0 +1,76 @@
+import carb.events
+import omni.kit.app as app
+import omni.ui as ui
+from typing import List, Optional
+from ..bindings import ICesiumOmniverseInterface
+
+
+class CesiumOmniverseStatisticsWidget(ui.Frame):
+    """
+    Widget that displays statistics about the scene.
+    """
+
+    NUMBER_OF_MATERIALS_LOADED_TEXT = "Number of materials loaded: {0}"
+    NUMBER_OF_GEOMETRIES_LOADED_TEXT = "Number of geometries loaded: {0}"
+    NUMBER_OF_GEOMETRIES_VISIBLE_TEXT = "Number of geometries visible: {0}"
+
+    def __init__(self, cesium_omniverse_interface: ICesiumOmniverseInterface, **kwargs):
+        super().__init__(build_fn=self._build_fn, **kwargs)
+
+        self._cesium_omniverse_interface = cesium_omniverse_interface
+
+        self._statistics_number_of_materials_loaded_field: Optional[ui.SimpleStringModel] = None
+        self._statistics_number_of_geometries_loaded_field: Optional[ui.SimpleStringModel] = None
+        self._statistics_number_of_geometries_visible_field: Optional[ui.SimpleStringModel] = None
+
+        self._subscriptions: List[carb.events.ISubscription] = []
+        self._setup_subscriptions()
+
+    def __del__(self):
+        self.destroy()
+
+    def destroy(self):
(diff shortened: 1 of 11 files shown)


exts/cesium.omniverse/cesium/omniverse/ui/statistics_widget.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class CesiumOmniverseStatisticsWidget(ui.Frame):
+    """
+    Widget that displays statistics about the scene.
+    """
+
+    NUMBER_OF_MATERIALS_LOADED_TEXT = "Number of materials loaded: {0}"
+    NUMBER_OF_GEOMETRIES_LOADED_TEXT = "Number of geometries loaded: {0}"
+    NUMBER_OF_GEOMETRIES_VISIBLE_TEXT = "Number of geometries visible: {0}"
```
weegeekps, line 15: Generally it's better to define static variables like this at the module level (outside of the class) rather than as a class variable. It hides them from any potential callers and ensures they'll only ever be interpreted once.

exts/cesium.omniverse/cesium/omniverse/ui/statistics_widget.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+
+    def _build_fn(self):
+        """Builds all UI components."""
+
+        with ui.VStack():
+            ui.Label("Statistics", height=0)
+            self._statistics_number_of_materials_loaded_field = ui.SimpleStringModel("")
+            self._statistics_number_of_geometries_loaded_field = ui.SimpleStringModel("")
+            self._statistics_number_of_geometries_visible_field = ui.SimpleStringModel("")
```
weegeekps, line 73: You should define the models in the `__init__` method. There's many reasons, including those outside of our control, where the `build_fn` could be called for a `Frame`. In those cases we'd recreate these and leak the old copies.

src/core/src/FabricUtil.cpp
```diff
@@ -497,4 +497,38 @@ std::string printFabricStage() {
...
+        auto paths = sip.getPathArray(geometryBuckets, bucketId);
+        statistics.numberOfGeometriesLoaded += paths.size();
+
+        auto worldVisibilityFabric =
+            sip.getAttributeArrayRd<bool>(geometryBuckets, bucketId, FabricTokens::_worldVisibility);
+        for (size_t i = 0; i < worldVisibilityFabric.size(); i++) {
+            if (worldVisibilityFabric[i]) {
+                statistics.numberOfGeometriesVisible++;
+            }
+        }
```
weegeekps, line 523: NIT: This seems like a good candidate for [std::accumulate](https://en.cppreference.com/w/cpp/algorithm/accumulate).

weegeekps (commented):
Looks good but I have a few comments.

src/core/src/FabricUtil.cpp
lilleyse: `std::count` also does the trick

exts/cesium.omniverse/cesium/omniverse/ui/statistics_widget.py
lilleyse: 👍 Also avoids having an optional type

weegeekps (approved):
Looks great.

PR #263, opened 2023-04-26
Fix type for tilesetId and tileId

lilleyse:
Copying over some changes from the [pool](https://github.com/CesiumGS/cesium-omniverse/tree/pool) branch:

* Fixes a mismatch between the Fabric type and the internal type. `tilesetId` and `tileId` should be `int64` everywhere.
* Adds some missing `const`


weegeekps (approved):
:shipit:

PR #264, opened 2023-04-26
GltfUtil cleanup

lilleyse:
Copying over some changes from the [pool](https://github.com/CesiumGS/cesium-omniverse/tree/pool) branch:

* Adds more helper functions to `GltfUtil`
  * `getDefaultBaseColorFactor`
  * `getDefaultMetallicFactor`
  * `getDefaultRoughnessFactor`
  * `hasNormals`
  * `hasTexcoords`
  * `hasImageryTexcoords`
  * `hasMaterial`
* Renamed `uv` to `texcoord` except when interacting with Fabric where the property is called `st`
* Ability to bake translation and scale into texcoords (used by the `pool` branch)

Issue #243, opened 2023-03-22
Issue watchlist for upcoming Kit releases


lilleyse:
This issue keeps tracks of things we're watching out for as new versions of Kit are released

* [x] Dynamic Texture Provider: https://github.com/CesiumGS/cesium-omniverse/issues/116
  * Fixed in Create 2022.3.3
  * Console warnings fixed in pre-release (4/22/2023)
* [x] Dynamic texcoords - for mesh pool
  * Fixed in pre-release (4/22/2023)
* [x] Orphaned prims: https://github.com/CesiumGS/cesium-omniverse/issues/214
  * Fixed in pre-release (3/28/2023)
* [x] Dynamic material support in Fabric: https://github.com/CesiumGS/cesium-omniverse/issues/242
  * `materialId` assignment fixed in pre-release (3/28/2023)
  * Dynamic material values are expected in Kit 105
* [x] Simplify prim deletion by removing [`/TempChangeTracking`](https://github.com/CesiumGS/cesium-omniverse/blob/0b2499e7faff26dbb357b0260af901e84c721254/src/core/src/FabricStageUtil.cpp#L934-L972) workaround
  * Not fixed as of pre-release (3/28/2023)
* [ ] USDRT material support and switch away from using Fabric directly.
  * Related: https://github.com/CesiumGS/cesium-omniverse/issues/94
* [x] Dynamic skies support in Fabric
  * Dragging in a sky works in Create 2022.3.3
  * Subsequent adjustments to time of day do not work (expected in Kit 105)
* [x] Fabric geometry doesn't become ready until camera stops moving: https://github.com/CesiumGS/cesium-omniverse/issues/255
  * [x] Fixed in pre-release (4/22/2023)
* [ ] Physics support in Fabric: https://github.com/CesiumGS/cesium-omniverse/issues/153
* [ ] Intermittent crashes and hangs (issue TBD)
* [x] `main` crashes in Linux in pre-release (5/1/2023)
* [ ] Custom materials don't get populated correctly when Fabric is enabled by default: https://github.com/CesiumGS/cesium-omniverse/pull/406
* [ ] Compressed texture fixes in Kit 105.1
* [ ] When custom material is saved `cesium.mdl` is converted to `./cesium.mdl`
* [ ] pipapi improvements: https://github.com/CesiumGS/cesium-omniverse/pull/383

PR #265, opened 2023-04-27
Geometry and material pool

### LINKED ISSUE #88: Tile flashing artifacts
When we are using `OmniPBR`, there is a brief flash of white while tiles are loading. This does not appear to be related to the synchronous material loads problem we were having earlier.

https://user-images.githubusercontent.com/915398/230966097-7b92544d-ce2b-46de-aa9a-ae5daa49c2f2.mp4

### LINKED ISSUE #255: Holes while camera is moving
Fabric prims don't finish loading until the camera stops moving, which results in holes.

https://user-images.githubusercontent.com/915398/230969896-43627ef8-a840-4905-a5ac-0638ad0bde61.mp4

<br>

I've tried a few different options to force synchronous loading, but no luck so far:

```
app.asyncRendering = false
app.usdrt.scene_delegate.geometryStreaming.enabled = false
rtx.hydra.geometrySyncLoads = true
```

lilleyse:
Fixes #88 

**Merge https://github.com/CesiumGS/cesium-omniverse/pull/252, https://github.com/CesiumGS/cesium-omniverse/pull/263, and https://github.com/CesiumGS/cesium-omniverse/pull/264 first.** This branch includes changes from those branches. Once those are merged I'm going to squash this into a single commit and mark as ready for review.

The main idea here is that we can preallocate a bunch of Fabric geometries and materials and reuse those as we stream in tiles. This has two main advantages:

* **Performance**: material compilation is slow and can hitch the main thread. By preallocating in advance we get better performance while navigating the scene. See [#169](https://github.com/CesiumGS/cesium-omniverse/pull/169) for why we can't just use a single material for all tiles (in short: texture asset paths aren't allowed as primvars)
* **Visual quality**: no more red flashes except when the tileset is initially loaded, since materials are now compiled in advance. There are still some white tiles that appear but whatever's causing this seems to have been fixed in the latest Kit pre-release.

Quick overview of the classes:

* **ObjectPool** - Generic pool class that doubles in capacity when it hits 75% occupancy. Objects in the pool are acquired and released as shared pointers.
* **FabricGeometryPool** - Subclass of `ObjectPool` that manages creation of `FabricGeometry` objects
* **FabricGeometry** - Encapsulates a Fabric geometry prim. Responsible for initializing, reusing, and destroying the geometry.
* **FabricGeometryDefinition** - A summary of a `FabricGeometry`. Geometries with different `FabricGeometryDefinition` can't be pooled together because the Fabric prim structure is different.
* **FabricMaterialPool** - Subclass of `ObjectPool` that manages creation of `FabricMaterial` objects. The initial capacity is 2000 which is roughly the number of tiles cesium-native has in memory when using the full default maximum memory usage for Cesium World Terrain. This seems to be the sweet spot of not being too slow when the tileset is first loaded (2-3 second stall) while avoiding having to expand the pool (and compile materials) at runtime.
* **FabricMaterial** - Encapsulates the various prims that make up a Fabric material. It also holds onto a `DynamicTextureProvider` whose image contents be swapped out whenever a new tile loads in. 
* **FabricMaterialDefinition** - Similar to `FabricGeometryDefinition` but extra strict because dynamic material values aren't supported in Kit 104.2 (e.g. tiles with `roughnessFactor` 0.0 and 1.0 will be placed in separate pools). This is also why translation and scale are baked into the texcoords, because every tile has different values for those and it would be impossible to reuse the same materials. With Kit 105 these restrictions will no longer apply.
* **FabricMeshManager** - Singleton and main entry point for acquiring and releasing `FabricMesh` objects
* **FabricMesh** - Encapsulates a `FabricGeometry` and `FabricMaterial`

**One last note** - geometry pooling is currently disabled on this branch for a combination of reasons:
* Texcoords can't be changed dynamically Kit 104.2 which is needed for geometry pooling. This has been fixed in a pre-release (see #243).
* Geometry pooling might be bad for performance since it means updating vertices frequently and potentially thrashing the BVH. This is just speculative at this point.

I was hoping to get geometry pooling working to fix #255, but we'll just need to wait until the next Kit release which fixes the underlying bug.


src/core/src/FabricMeshManager.cpp
```diff
@@ -0,0 +1,213 @@
...
+    releaseGeometry(geometry);
+
+    if (material != nullptr) {
+        releaseMaterial(material);
+    }
+}
+
+uint64_t FabricMeshManager::getNumberOfGeometriesInUse() const {
+    uint64_t count = 0;
+    for (const auto geometryPool : _geometryPools) {
```
weegeekps, line 75: All of these loops perform copies. These should be references.

```suggestion
    for (const auto &geometryPool : _geometryPools) {
```

src/core/src/FabricMaterial.cpp
```diff
@@ -0,0 +1,351 @@
...
+        attributes.addAttribute(FabricTypes::_cesium_tileId, FabricTokens::_cesium_tileId);
+        // clang-format on
+
+        if (hasBaseColorTexture) {
+            attributes.addAttribute(FabricTypes::diffuse_texture, FabricTokens::diffuse_texture);
+        }
+
+        attributes.createAttributes(shaderPathFabric);
+
+        const size_t paramColorSpaceSize = hasBaseColorTexture ? 2 : 0;
```
weegeekps, line 179: You may want to add some comments here to explain what all the magic numbers are. It may be obvious now, but it may not be obvious in 6+ months.

weegeekps (commented):
Some minor comments.

src/core/include/cesium/omniverse/ObjectPool.h
```diff
@@ -0,0 +1,88 @@
...
+  public:
+    ObjectPool() {}
+
+    virtual ~ObjectPool() = default;
+
+    std::shared_ptr<T> acquire() {
+        const auto percentActive = computePercentActive();
+
+        if (percentActive > _doublingThreshold) {
+            const auto newCapacity = std::max(_capacity * 2, uint64_t(1));
```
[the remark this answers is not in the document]
lilleyse: `_capacity` is initially 0, so it's to safeguard against that. I added a comment to clarify.

lilleyse:
@weegeekps noticed that there's a segfault when adding Cesium World Terrain (blank) then Bing Maps Aerial to it. I still need to look into that.

weegeekps (approved):
Since #270 is where that segfault is getting fixed, I'm comfortable going ahead and approving this PR now as it looks good to go.

Issue #258, opened 2023-04-19
Additional stage cameras stop loding tiles after movement


weegeekps:
Reported by the NVIDIA team:

> The perspective camera works fine, but adding a stage camera that can be keyframed in the sequencer stops loading tiles after some movement.

This is going to need some digging to solve, but the current hypothesis is that we are only paying attention to one camera for updates, rather than whatever the user's selected camera is.

lilleyse:
Also broken for multi-viewport setups


https://user-images.githubusercontent.com/915398/234914853-c3bf1be2-b25c-418b-b947-060a50de00f4.mp4

lilleyse:
Cesium Native supports multiple view states, we just need to fix the plumbing in our extension.

To get the list of viewports use

```
from omni.kit.viewport.window import get_viewport_window_instances
viewports = get_viewport_window_instances()
```

And then get the `viewport_api` attribute to access the projection matrix.

PR #268, opened 2023-04-27
Add support for multiple viewports

lilleyse:
Addresses https://github.com/CesiumGS/cesium-omniverse/issues/258#issuecomment-1525926299

`onUpdateFrame` now takes a vector of `Viewport` objects which are passed along to Cesium Native.

https://user-images.githubusercontent.com/915398/234978409-db538b52-d7ac-4dd3-8679-b8db6f361a80.mp4


weegeekps (approved):
This works well. One thing it made me realize is that the credit system is not applying the credits to every viewport, so I opened https://github.com/CesiumGS/cesium-omniverse/issues/269 to resolve this.

PR #270, opened 2023-05-01
Fix adding imagery to existing tileset

lilleyse:
Fixes a bug introduced in https://github.com/CesiumGS/cesium-omniverse/pull/204 when adding imagery to an existing tileset. I thought this bug was fixed already in another PR but I guess not.


weegeekps (approved):
:shipit:

PR #272, opened 2023-05-02
Only select tiles when tileset is visible

lilleyse:
While it could be useful to load tiles when a tileset is hidden (see [`preloadTilesWhenHidden`](https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTileset.html#preloadWhenHidden) in CesiumJS), this shouldn't be the default behavior. Another positive side effect of this fix is that credits aren't show when the tileset is hidden.


weegeekps (approved):
LGTM

PR #273, opened 2023-05-02
Enable async rendering in cesium.omniverse.app.kit

lilleyse:
Originally we needed to set `app.asyncRendering = false` to work with Fabric, but this is no longer required.

`asyncRendering` is true by default so we can just remove the entry from `cesium.omniverse.app.kit`.


weegeekps (approved):
🚢

PR #274, opened 2023-05-03
Fix freeing of `pLoadThreadResult`

lilleyse:
When a tile is freed before it's prepared in the main thread we should be releasing the fabric meshes back into the pool before destroying the tile. The fix is to free `pLoadThreadResult` in the same way that `pMainThreadResult` is freed.


weegeekps (approved):
:shipit:

PR #275, opened 2023-05-03
Added packman user files to the .gitignore file

weegeekps:
Packman's documentation states:

> User files are created to override settings in project files in conjunction with the pull command and the import element in project file. This is useful for local experimentation, for instance when using packages built from source instead of those pulled from a packman remote.

Turns out these are super useful when testing different versions of kit other than what we currently are targeting. I'm adding the extension to the `.gitignore` file so we don't have to remove these when dealing with prerelease versions of packages between branches.


lilleyse:
@weegeekps awesome, this is super useful since I'm constantly switching between kit versions.

Could you add some instructions to the developer setup guide for how to use `.user` files?

* Where to put the `.user` file
* An example `.user` file that uses a different Kit version

lilleyse:
The example works nicely

PR #276, opened 2023-05-03
Add back `wrap_clamp` texture wrapping

lilleyse:
Fixes a regression introduced in https://github.com/CesiumGS/cesium-omniverse/pull/265 where texture clamping was no longer working. The fix is to add the `lookup_color` and `texture_coordinate_2d` nodes like we were doing in [FabricStageUtil.cpp](https://github.com/CesiumGS/cesium-omniverse/blob/8832399c0d434f28ba7e9c2cd5c322099b64f4cc/src/core/src/FabricStageUtil.cpp).

Click the images to see the difference more clearly:

Before|After
--|--
before|after


weegeekps (approved):
LGTM

PR #278, opened 2023-05-05
Added premake5.lua files to better support Kit templates

weegeekps:
These are used by several of the Kit templates, including the Kit Application template. This way installing the extension into a Kit app template is as simple as dragging and dropping our extensions into the extensions folder.


lilleyse:
Looks good, no need to wait for CI for this one.

PR #280, opened 2023-05-12
Add more render statistics

lilleyse:
Added some more render statistics to help with debugging:

* Number of triangles loaded
* Number of triangles visible
* Tileset cached bytes

I eventually want to add "Number of texels loaded" and "Number of texels visible" but it isn't as easy to add right now due to how things are structured.

image

### DIFF
--- a/exts/cesium.omniverse/cesium/omniverse/ui/debug_window.py
+++ b/exts/cesium.omniverse/cesium/omniverse/ui/debug_window.py
@@ -17,7 +17,7 @@ class CesiumOmniverseDebugWindow(ui.Window):
 
         self._logger = logging.getLogger(__name__)
         self._cesium_omniverse_interface = cesium_omniverse_interface
-        self._cesium_message_field: Optional[ui.SimpleStringModel] = None
+        self._cesium_message_field: ui.SimpleStringModel = ui.SimpleStringModel("")
         self._statistics_widget: Optional[CesiumOmniverseStatisticsWidget] = None
 
         # Set the function that is called to build widgets when the window is visible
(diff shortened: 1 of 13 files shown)


exts/cesium.omniverse/cesium/omniverse/ui/debug_window.py
```diff
@@ -60,10 +60,8 @@ def print_fabric_stage():
...
         with ui.VStack():
-            ui.Button("Remove all Tilesets", clicked_fn=lambda: remove_all_tilesets())
-            ui.Button("Reload all Tilesets", clicked_fn=lambda: reload_all_tilesets())
-            ui.Button("Print Fabric stage", clicked_fn=lambda: print_fabric_stage())
-            with ui.VStack():
-                self._cesium_message_field = ui.SimpleStringModel("")
-                ui.StringField(self._cesium_message_field, multiline=True, read_only=True)
+            ui.Button("Remove all Tilesets", height=20, clicked_fn=lambda: remove_all_tilesets())
+            ui.Button("Reload all Tilesets", height=20, clicked_fn=lambda: reload_all_tilesets())
+            ui.Button("Print Fabric stage", height=20, clicked_fn=lambda: print_fabric_stage())
```
weegeekps, line 65: Can you go ahead and fix all of these? We can just pass in the functions as higher-order and this should work. You may need to stub out some parameters using `_` but I don't think you will in this case.

exts/cesium.omniverse/cesium/omniverse/ui/statistics_widget.py
```diff
@@ -46,7 +52,8 @@ def _on_update_frame(self, _e: carb.events.IEvent):
         if not self.visible:
             return
 
-        fabric_statistics = self._cesium_omniverse_interface.get_fabric_statistics()
+        render_statistics = self._cesium_omniverse_interface.get_render_statistics()
+        fabric_statistics = render_statistics.fabric_statistics
```
weegeekps, line 56: NIT: I have mixed feelings here. There's little reason to pull the `fabric_statistics` property out here into it's own variable. It doesn't help readability that much and I think an argument can be made that we're hiding that the fabric statistics are part of render_statistics.

On the other hand... :shrug:

weegeekps (commented):
One major comment to consider.

exts/cesium.omniverse/cesium/omniverse/ui/statistics_widget.py
lilleyse: I was on the fence about this too. I was trying to simplify the C++ side but I think a flat struct is cleaner.

weegeekps (approved):
LGTM

PR #284, opened 2023-05-16
fix-carbonite-import-error

corybarr:
It looks like NVIDIA has moved the location of the Carbonite SDK since I joined and installed Omniverse. extension.py has this issues:

image

This should fix it.


lilleyse:
Thanks @corybarr!

lilleyse:
There were more structural changes in Kit 105 so I regenerated both the window and linux `.code-workspace` files. That's currently sitting in the internal Kit 105 branch.

PR #286, opened 2023-05-17
Add debug options to Cesium Data prim

lilleyse:
This will help us tweak common debug options without needing to recompile the code.

image


exts/cesium.omniverse/cesium/omniverse/ui/attributes/data_attributes_widget.py
```diff
@@ -23,5 +23,11 @@ def _customize_props_layout(self, props):
                 CustomLayoutProperty("cesium:georeferenceOrigin:latitude", "Latitude")
                 CustomLayoutProperty("cesium:georeferenceOrigin:longitude", "Longitude")
                 CustomLayoutProperty("cesium:georeferenceOrigin:height", "Height")
+            with CustomLayoutGroup("Debug Options"):
```
weegeekps, line 26: Let's start this collapsed. Just add `collapsed=True` to the arguments for `CustomLayoutGroup`.

exts/cesium.usd.plugins/schemas/cesium_schemas.usda
```diff
@@ -63,6 +63,47 @@ class CesiumDataPrim "CesiumDataPrim" (
         displayName = "Georeference Origin Height"
         doc = "The height of the origin in meters above the WGS84 ellipsoid. Do not confuse this with a geoid height or height above mean sea level, which can be tens of meters higher or lower depending on where in the world the origin is located."
     )
+
+    bool cesium:debug:disableMaterials = false (
+        customData = {
+            string apiName = "debugDisableMaterials"
+        }
+        displayName = "Disable Materials"
+        doc = "For debugging only: whether to disable materials."
```
weegeekps, line 72: I'm not a fan of how these are written. To a native English speaker it will probably be okay, but to a non-native speaker I can see this adding confusion. Complete sentences such as, "Debugging value which disables materials when enabled," would be better.

src/core/src/Context.cpp
```diff
@@ -231,6 +233,13 @@ void Context::processPropertyChanged(const ChangedPrim& changedPrim) {
                     tilesetsToReload.emplace(tileset);
                 }
             }
+        } else if (
```
weegeekps, line 236: Lots of conditional nesting here. I think the outer conditional is a good candidate for a switch with branching functions for readability, as long as the additional call doesn't cause too much overhead.

weegeekps (commented):
Looking good but some initial thoughts to make this cleaner.

weegeekps (approved):
Oops, meant to approve this yesterday and didn't click submit. LGTM.

PR #288, opened 2023-05-18
Added regenerate_schema.bat for Windows

lilleyse:
I was getting tired of switching to Linux to regenerate schemas so I wrote a batch file for Windows. I also removed some of the manual steps we needed to do every time.

* Previously `module.cpp` and `moduleDeps.cpp` would be empty after running `regenerate_schema.sh`. This is because they were copied to the folder before the step that removes the first 23 lines from each file. Fixed by copying them after.
* Previously we would need to revert changes to `plugInfo.json.in`. This is no longer required, instead we can set `PLUG_INFO_RESOURCE_PATH ` and `PLUG_INFO_ROOT ` in CMake. I haven't noticed any issues now that `Name` is `cesium` instead of `CesiumUsdSchemas`. (**EDIT: I reverted the name change, we still need to manually rename cesium to CesiumUsdSchemas**)

We still have to manually add `CESIUM_API` to the class definition and remove it from the function signatures. Interestingly, Windows works totally fine if we don't do this... it's Linux that crashes at startup.


weegeekps (approved):
Sorry, I thought I had already approved this.

PR #290, opened 2023-05-19
Fixes credits not appearing on all viewports

### LINKED ISSUE #269: Credits are not applied to every viewport
Currently we only apply credits to the primary viewport, but we should be applying them to every viewport.

corybarr:
Resolves #269 

Adds credits to multiple viewpoints:

image


exts/cesium.omniverse/cesium/omniverse/extension.py
```diff
@@ -217,6 +218,11 @@ def _on_update_frame(self, _):
             viewport.height = float(viewport_api.resolution[1])
             viewports.append(viewport)
 
+        if len(viewports) != self._num_credits_viewport_frames:
+            self._logger.info("CreditViewportFrame: num viewports changed, triggering CreditsViewportFrames setup")
```
weegeekps, line 222: I know we're pretty inconsistent about this, but it would be better to log this as a debugging message rather than as an informational one, since it's in the frame update.

exts/cesium.omniverse/cesium/omniverse/ui/credits_viewport_controller.py
```diff
@@ -0,0 +1,89 @@
...
+class CreditsViewportController:
+    _instance = None
+    _cesium_omniverse_interface = None
+    _some_string = None
+    _logger: Optional[logging.Logger] = None
+    _parsed_credits: List[ParsedCredit] = []
+    _credits: List[Tuple[str, bool]] = []
+    _event_handlers = []
+    _subscriptions: List[carb.events.ISubscription] = []
+    _initted: bool = False
```
weegeekps, line 19: So, Python can be a bit weird. All of these should be instantiated in the `__init__` block, and all `None` values should have a type hint defined.

The way this is done right now will actually declare these as a class variable rather than an instance variable. While on the surface this seems the same, class variables actually share copies between every instance of the class when the class is first initialized, and instance variables are unique values for every class. When you're dealing with arrays and such, this can be a huge footgun if you're not careful. Since this controller is basically acting as a singleton right now we're not running into issues but we should absolutely move these declarations into the `__init__`.

exts/cesium.omniverse/cesium/omniverse/extension.py
```diff
@@ -52,7 +53,7 @@ def __init__(self) -> None:
         self._main_window: Optional[CesiumOmniverseMainWindow] = None
         self._asset_window: Optional[CesiumOmniverseAssetWindow] = None
         self._debug_window: Optional[CesiumOmniverseDebugWindow] = None
-        self._credits_viewport_frame: Optional[CesiumCreditsViewportFrame] = None
+        self._credits_viewport_frames: List[Optional[CesiumCreditsViewportFrame]] = []
```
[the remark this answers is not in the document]
corybarr: Makes sense. Changed.

exts/cesium.omniverse/cesium/omniverse/extension.py
corybarr: True. That would be better. What if we just ditch it? I don't see it being of any use to an end user.

exts/cesium.omniverse/cesium/omniverse/extension.py
```diff
@@ -407,3 +413,18 @@ def show_debug_window(self, _menu, value):
...
         elif self._debug_window is not None:
             self._debug_window.visible = False
+
+    def _setup_credits_viewport_frames(self):
+        self._destroy_credits_viewport_frames()
+        viewport_frames = []
+        for instance in get_viewport_window_instances():
+            credits_viewport_frame = CesiumCreditsViewportFrame(_cesium_omniverse_interface, instance)
+            viewport_frames.append(credits_viewport_frame)
+        self._credits_viewport_frames = viewport_frames
```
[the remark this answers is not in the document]
corybarr: True. Changed.

exts/cesium.omniverse/cesium/omniverse/extension.py
```diff
@@ -407,3 +413,18 @@ def show_debug_window(self, _menu, value):
...
+
+    def _destroy_credits_viewport_frames(self):
+        for credits_viewport_frame in self._credits_viewport_frames:
+            if credits_viewport_frame is not None:
+                credits_viewport_frame.destroy()
+                credits_viewport_frame = None
+        CreditsViewportController().clear_handlers()
```
corybarr, line 430: @weegeekps Should clear_handlers have a classmethod decorator so the call is `CreditsViewportController.clear_handlers()`?

exts/cesium.omniverse/cesium/omniverse/extension.py
```diff
@@ -407,3 +413,18 @@ def show_debug_window(self, _menu, value):
...
+        for instance in get_viewport_window_instances():
+            credits_viewport_frame = CesiumCreditsViewportFrame(_cesium_omniverse_interface, instance)
+            viewport_frames.append(credits_viewport_frame)
+        self._credits_viewport_frames = viewport_frames
+
+    def _destroy_credits_viewport_frames(self):
+        for credits_viewport_frame in self._credits_viewport_frames:
+            if credits_viewport_frame is not None:
+                credits_viewport_frame.destroy()
+                credits_viewport_frame = None
```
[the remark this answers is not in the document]
corybarr: Makes sense. Changed.

exts/cesium.omniverse/cesium/omniverse/ui/credits_viewport_frame.py
```diff
@@ -71,13 +74,12 @@ def _on_update_frame(self, _e: carb.events.IEvent):
...
             self._data_attribution_button.visible = has_offscreen_credits
 
-        self._cesium_omniverse_interface.credits_start_next_frame()
-
     def _on_data_attribution_button_clicked(self):
         self._credits_window = CesiumOmniverseCreditsWindow(self._cesium_omniverse_interface, self._credits)
 
     def _build_fn(self):
         with self._credits_viewport_frame:
+            self._logger.info("CreditViewportFrame: rebuilding")
```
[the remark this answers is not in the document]
corybarr: True. Gone.

exts/cesium.omniverse/cesium/omniverse/extension.py
weegeekps: I'm fine with ditching it entirely.

exts/cesium.omniverse/cesium/omniverse/extension.py
weegeekps: Probably not given the move to making it less a pure singleton, but let me mull over it some and get back to you in the morning.

exts/cesium.omniverse/cesium/omniverse/ui/credits_viewport_controller.py
corybarr: I actually moved these out of __init__ so they'd be class variables because I'm thinking like a C#/++ singleton person. But sounds like it's more Pythonic to use a global object, so--yeah--let me move the approach to that.

exts/cesium.omniverse/cesium/omniverse/ui/credits_viewport_controller.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def __new__(cls, *args, **kwargs):
+        if not cls._instance:
+            cls._instance = super().__new__(cls, *args, **kwargs)
+        return cls._instance
```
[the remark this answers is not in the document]
corybarr: >Let me know if you want to meet and I can pair with you and clarify anything if you have any questions.

Yeah, let's grab some time maybe tomorrow. I'm curious about the potential future problems due to hot-loading. And there's some future cases I'd like to mull over. Like what if 2+ extensions want to customize viewport visualization? For example, maybe one extension controls shading for visualization of different semantic sections while another does something else?

exts/cesium.omniverse/cesium/omniverse/ui/credits_viewport_frame.py
```diff
@@ -99,3 +101,9 @@ def _build_fn(self):
                         height=0,
                         clicked_fn=self._on_data_attribution_button_clicked,
                     )
+
+    def set_new_credits(self, credits):
+        self._new_credits = credits
+
+    def handle_event(self, event):
+        self._new_credits = event
```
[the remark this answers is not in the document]
corybarr: set_new_credits is unused. It's now gone. For a quick first pass I was using a setter instead of an event. I forgot to delete the setter.

My old IDE (Visual Studio) would flag unused methods (in C#/C++). I'd like to find a way to get VS Code to do that for Python.

exts/cesium.omniverse/cesium/omniverse/ui/credits_viewport_controller.py
```diff
@@ -0,0 +1,89 @@
+from .credits_parser import CesiumCreditsParser, ParsedCredit
+from typing import List, Optional, Tuple
+import logging
+import carb.events
+import omni.kit.app as app
+from ..bindings import ICesiumOmniverseInterface
+import omni.ui as ui
+
+
+class CreditsViewportController:
```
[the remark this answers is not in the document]
corybarr: added

exts/cesium.omniverse/cesium/omniverse/extension.py
corybarr: @weegeekps The controller's been refactored based on our discussion about not using a singleton.

exts/cesium.omniverse/cesium/omniverse/extension.py
```diff
@@ -407,3 +412,16 @@ def show_debug_window(self, _menu, value):
...
+        self._credits_viewport_frames = [
+            CesiumCreditsViewportFrame(_cesium_omniverse_interface, i) for i in get_viewport_window_instances()
+        ]
+        if self._credits_viewport_controller is not None:
+            self._credits_viewport_controller.broadcast_credits()
+
+    def _destroy_credits_viewport_frames(self):
+        for credits_viewport_frame in self._credits_viewport_frames:
+            credits_viewport_frame.destroy()
+        self._credits_viewport_frames.clear()
```
weegeekps, line 427: Potential Nitpick: Do you expect to call these multiple times? If not, I'm not sure it's worth breaking them out into separate methods. I can see either way though.

exts/cesium.omniverse/cesium/omniverse/ui/credits_viewport_controller.py
```diff
@@ -0,0 +1,70 @@
...
+import omni.ui as ui
+import omni.kit.app
+import json
+from carb.events import IEventStream
+
+
+class CreditsViewportController:
+    def __init__(self, cesium_omniverse_interface: ICesiumOmniverseInterface):
+        self._cesium_omniverse_interface = cesium_omniverse_interface
+        self._logger: Optional[logging.Logger] = None
```
weegeekps, line 16: No need to declare and set this here, you're setting up the logger below.

weegeekps (commented):
Looks really good. Just a few more small comments and I think this will be good to go.

exts/cesium.omniverse/cesium/omniverse/extension.py
corybarr: I don't expect to call these multiple times in the short term, but probably in the long term as we develop the extension to handle multiple viewports that can be setup, configured, and destroyed on the fly.

Thinking about it, this method should be in the controller. Perhaps it's best to push that task and this question into the larger plan of modularizing extension.py?

weegeekps: > Perhaps it's best to push that task and this question into the larger plan of modularizing extension.py?

Yeah, I'm cool with that.

exts/cesium.omniverse/cesium/omniverse/ui/credits_viewport_controller.py
corybarr: Ah, I had `__init__` set up where everything is declared as None or [] with type hinting in a block then instantiated later. Not sure where I got that pattern. No need to use it.

Changed.

exts/cesium.omniverse/cesium/omniverse/ui/credits_viewport_controller.py
```diff
@@ -0,0 +1,70 @@
...
+        self._parsed_credits: List[ParsedCredit] = []
+        self._credits: List[Tuple[str, bool]] = []
+        self._subscriptions: List[carb.events.ISubscription] = []
+
+        self._setup_update_subscription()
+        self._message_bus: IEventStream = omni.kit.app.get_app().get_message_bus_event_stream()
+        self._logger = logging.getLogger(__name__)
+        self._EVENT_CREDITS_CHANGED: int = carb.events.type_from_string(
+            "cesium.omniverse.viewport.ON_CREDITS_CHANGED"
+        )
```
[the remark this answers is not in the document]
corybarr: Do we want to create something like [constants.py](https://github.com/CesiumGS/cesium-omniverse-demos/blob/main/source/extensions/cesium.ebc_demo.window/cesium/ebc_demo/window/constants.py)? Both the credits_viewport_frame and credits_viewport_controller need access, and that seems much better than typing in an error-prone string.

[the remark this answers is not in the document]
corybarr: And, if so, would it be an extension-wide file to put event constants?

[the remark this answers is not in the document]
corybarr: added events.py

weegeekps (approved):
🚀

PR #292, opened 2023-05-22
Force extension reload

### LINKED ISSUE #151: Reload application when extension is added for the first time
Is it possible for an extension to prompt the user to reload the application when enabled for the first time? Our extension needs to be able to register some custom USD plugins at startup.

Similar to what you might see in VS Code:

image

corybarr:
Resolves #151 

Prevents the extension from running without setting it to Autoload and restarting a Kit app. If trying to live-enable the extension, users will get this modal:

image


corybarr:
> Thanks @corybarr, nice and simple!
> 
> * I think we can remove this line from the README now:
>   > **NOTE: Once enabling this extension via "Autoload," you must restart Omniverse for it to load the necessary USD components.**
> * Could you work with @r-veenstra to update the [quick start tutorial](https://cesium.com/learn/omniverse/omniverse-quickstart/) after the June 1st release?

@lilleyse Sure, will sync to coord the quickstart.

lilleyse:
I pushed the README change since it was small: https://github.com/CesiumGS/cesium-omniverse/pull/292/commits/aefb09a91d90b32f8fc45145f8ddbc0005f77057

Thanks @corybarr!

PR #293, opened 2023-05-22
Set black to Python formatter

### LINKED ISSUE #291: Make black the default Python formatter in the workspace file
With black the default Python formatter in the workspace file, issues that would cause CI to fail should be visible before committing. 

image

corybarr:
Resolve #291 

This sets the default Python interpret to Black in the VS Code workspace files. It will highlight more formatting issues at code time that could cause continuous integration to fail.

### DIFF
--- a/.vscode/cesium-omniverse-linux.code-workspace
+++ b/.vscode/cesium-omniverse-linux.code-workspace
@@ -55,7 +55,8 @@
       "editorconfig.editorconfig",
       "cheshirekow.cmake-format",
       "redhat.vscode-yaml",
-      "tamasfe.even-better-toml"
+      "tamasfe.even-better-toml",
+      "ms-python.black-formatter"
     ],
     "unwantedRecommendations": []
   }
--- a/.vscode/cesium-omniverse-linux.code-workspace
+++ b/.vscode/cesium-omniverse-linux.code-workspace
@@ -42,6 +42,7 @@
     "python.linting.flake8Enabled": true,
     "[python]": {
       "editor.defaultFormatter": "ms-python.black-formatter"
+    }
   },
   "extensions": {
     "recommendations": [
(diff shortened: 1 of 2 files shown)


.vscode/cesium-omniverse-linux.code-workspace
```diff
@@ -39,7 +39,9 @@
     "python.analysis.typeCheckingMode": "basic",
     "python.linting.enabled": true,
     "python.linting.pylintEnabled": false,
-    "python.linting.flake8Enabled": true
+    "python.linting.flake8Enabled": true,
+    "[python]": {
+      "editor.defaultFormatter": "ms-python.black-formatter"
   },
```
lilleyse, line 45: Missing a closing `}`

.vscode/cesium-omniverse-windows.code-workspace
```diff
@@ -427,7 +427,10 @@
     "python.analysis.typeCheckingMode": "basic",
     "python.linting.enabled": true,
     "python.linting.pylintEnabled": false,
-    "python.linting.flake8Enabled": true
+    "python.linting.flake8Enabled": true,
+    "[python]": {
+      "editor.defaultFormatter": "ms-python.black-formatter"
```
lilleyse, line 432: `ms-python.black-formatter` should be added to the extension recommendations list below

lilleyse (changes requested):

.vscode/cesium-omniverse-windows.code-workspace
corybarr: Good call. Added.

.vscode/cesium-omniverse-linux.code-workspace
corybarr: Added

lilleyse (approved):

PR #299, opened 2023-05-26
Separate tile loading into stages

lilleyse:
Separates tile loading into three stages:

1. Gather meshes from the glTF
2. Acquire Fabric prims from the pool
3. Set mesh data on the Fabric prims

The eventual goal is to do step 1 in the load thread, step 2 in the main thread, and step 3 back in the load thread. The is based on the recommendation that Fabric topology edits should happen on the main thread (topology edits being prim or attribute creation/destruction) while value updates could happen on the load thread.

Unfortunately step 3 is not working in the load thread yet, but to see what that looks like revert https://github.com/CesiumGS/cesium-omniverse/pull/299/commits/b2b0e078e78f66096bfa6a54fe86a2252e7cc893.

### DIFF
--- a/src/core/include/cesium/omniverse/FabricMesh.h
+++ b/src/core/include/cesium/omniverse/FabricMesh.h
@@ -21,6 +21,20 @@ class FabricMesh {
 
     void setVisibility(bool visible) const;
 
+    void setTile(
+        int64_t tilesetId,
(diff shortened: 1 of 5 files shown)


weegeekps (approved):
This looks fine to me.

Issue #282, opened 2023-05-15
RTX real-time lighting issues with some tilesets


r-veenstra:
Some 3D tilesets are overly dark when rendered with RTX real-time lighting.  This isn't consistent across all datasets, so it appears to be dataset specific.  See below.

Google 3D tiles - render issues with RTX Real-time
![Google_Lighting](https://github.com/CesiumGS/cesium-omniverse/assets/123468416/c813a25c-4eeb-4c92-95eb-f7c2bdf90ae4)

Aerometrex Denver - no issues
![AMX_Lighting](https://github.com/CesiumGS/cesium-omniverse/assets/123468416/e278e2e9-d858-4bd6-8676-32d6b5ebd10a)

Melbourne Photogrammetry - render issues with RTX Real-time
![Melbourne Photogrammetry](https://github.com/CesiumGS/cesium-omniverse/assets/123468416/b3ce7b65-7536-4b43-aaa6-2ea474307b00)

lilleyse:
This also seems to be related to the HDRI lighting rig in cesium-omniverse-samples because the default lighting looks fine.

![image](https://github.com/CesiumGS/cesium-omniverse/assets/915398/eb3af3cd-a77e-4476-afd6-95ab0751a8ce)

r-veenstra:
Interesting.  I'll do some experimentation with some different light types and configuration.

r-veenstra:
@lilleyse the issue is still there, and it appears specific to any HDR on a Dome Light, including the stock stage lighting.

The reason it is less visible on the stock stage is:

1) The stock stage uses a directional light for the sun, which does not appear to be impacted in RTX Real Time
2) The stock HDRI Dome light is just for background/bounce light and is relatively low intensity compared to the other HDRI's, including NVIDIA's stock ones, which have patches of very high sunlight intensity built in.

If I reduce the stock directional lights intensity to zero, and just rely on the stock dome light, you'll see the RTX version is still quite a bit darker.

![image](https://github.com/CesiumGS/cesium-omniverse/assets/123468416/648b4e65-fbb3-4f22-9f49-475f1d37bb9a)

My next step will be to check out the differences from a material point of view between datasets that are successful and those that are not.  Given the Photorealistic tiles are designed to be rendered unlit, it would not surprise me if they have some odd stock settings that cause a problem.

kring:
[...]

Issue #36, opened 2022-11-03
Handle unlit materials


lilleyse:
If a glTF uses the `KHR_materials_unlit` extension we should render with an unlit material and not generate normals.

We could also give some control to the user. There could be a drop-down list to override the default behavior. The options would be UNLIT, LIT, or AUTO.

lilleyse:
Cesium for Unreal supports unlit materials by disabling shadows and making all normals point up (along the ellipsoid surface normal: https://github.com/CesiumGS/cesium-unreal/pull/1017

I tried experimenting with the same idea in Omniverse and the results don't look great. Aside from ray self-intersections which show up as black, there's this flat shading type of look regardless of the normal direction. It reminds me of some of the issues in https://github.com/CesiumGS/cesium-omniverse/issues/59.

Smooth normals|Flat normals|Up normals
--|--|--
![smooth-normals](https://github.com/CesiumGS/cesium-omniverse/assets/915398/5cc9386c-89c6-412d-8f0f-378e8fc020be)|![flat-normals](https://github.com/CesiumGS/cesium-omniverse/assets/915398/0f241032-1d68-4c4e-ae36-5e86461d0f73)|![wgs84-normals](https://github.com/CesiumGS/cesium-omniverse/assets/915398/4b5933d8-3f32-4c0f-9776-da4bee8315b3)

[unlit-normals-test.usda.zip](https://github.com/CesiumGS/cesium-omniverse/files/11567402/unlit-normals-test.usda.zip)

Here's the code:

```c++
pxr::VtArray<pxr::GfVec3f> getNormalsGlobe(
    const pxr::VtArray<pxr::GfVec3f>& positions,
    const glm::dmat4& localToEcefTransform) {

    const auto ecefToLocalTransform = glm::dmat3(glm::inverse(localToEcefTransform));

    pxr::VtArray<pxr::GfVec3f> normalsUsd(positions.size(), pxr::GfVec3f(0.0f));

    for (size_t i = 0; i < positions.size(); i++) {
        const auto position = UsdUtil::usdToGlmVector(positions[i]);
        const auto positionEcef = glm::dvec3(localToEcefTransform * glm::dvec4(position, 1.0));
        const auto normalEcef = glm::normalize(positionEcef); // geocentric normal, should be geodetic
        const auto normal = ecefToLocalTransform * normalEcef;
[...]

PR #300, opened 2023-05-26
Set better default values for unlit materials

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/282

Sets metallicFactor to 0.0 and roughnessFactor to 1.0 for materials with the KHR_materials_unlit extension.

This improves the look of Google 3D Tiles until we're able to properly support unlit materials: #36


weegeekps (commented):
:rocket:

weegeekps (approved):

PR #302, opened 2023-05-30
Change material loading color

lilleyse:
When materials are compiling we draw a solid red color. I changed it to white which is a bit less harsh looking.

I was also considering black because it blends with the empty background better. Or cyan because that's what Cesium for Unity does. If anyone has any strong preferences I'd be happy to change it.

### DIFF
--- a/src/core/src/FabricGeometry.cpp
+++ b/src/core/src/FabricGeometry.cpp
@@ -20,7 +20,7 @@ namespace cesium::omniverse {
 
 namespace {
 
-const auto MATERIAL_LOADING_COLOR = pxr::GfVec3f(1.0f, 1.0f, 1.0f);
+const auto MATERIAL_LOADING_COLOR = pxr::GfVec3f(0.0f, 0.0f, 0.0f);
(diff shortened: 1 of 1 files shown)


corybarr:
> When materials are compiling we draw a solid red color. I changed it to white which is a bit less harsh looking.
> 
> I was also considering black because it blends with the empty background better. Or cyan because that's what Cesium for Unity does. If anyone has any strong preferences I'd be happy to change it.

I don't have a strong preference, but if I had to choose I'd go with black to hide the issue.

weegeekps:
+1 for the black vote. I know that the Nvidia guys suggested that as well.

lilleyse:
Ok, I changed it to black.

weegeekps (approved):
Looks fine. It might be worthwhile to make this configurable through options exposed in the Cesium prim as a debugging option. Perhaps a future ticket?

lilleyse:
Sure, that could be a separate ticket. Maybe a good one for Matt?

PR #304, opened 2023-05-30
Use Cesium Native's performance tracing utility

lilleyse:
Adds a new CMake option to enable performance tracing with Cesium Native: `CESIUM_OMNI_ENABLE_TRACING`

```sh
cmake -B build -D CESIUM_OMNI_ENABLE_TRACING
cmake --build build
```

A file called `cesium-trace-xxxxxxxxxxx.json` will be saved to the `exts/cesium-omniverse` folder when the program exits. This file can then be inspected in `chrome://tracing`.

Note that the JSON output may get truncated if the program closes unexpectedly - e.g. when the debugging session is stopped or the program crashes - or if `app.fastShutdown` is `true` (like with Omniverse Create). Therefore the best workflow for performance tracing is to run `cesium.omniverse.app.kit` and close the window normally.

image


src/core/src/Context.cpp
```diff
@@ -90,10 +94,20 @@ void Context::initialize(int64_t contextId, const std::filesystem::path& cesiumE
...
 
     Cesium3DTilesSelection::registerAllTileContentTypes();
+
+    CESIUM_TRACE_INIT((cesiumExtensionLocation /
+                       fmt::format(
+                           "cesium-trace-{}.json",
+                           std::chrono::time_point_cast<std::chrono::microseconds>(std::chrono::steady_clock::now())
+                               .time_since_epoch()
+                               .count()))
+                          .string());
```
lilleyse, line 104: Alternatively this could be written as:

```c++
#if CESIUM_TRACING_ENABLED
    const auto timestamp = std::chrono::time_point_cast<std::chrono::microseconds>(std::chrono::steady_clock::now())
                               .time_since_epoch()
                               .count();
    const auto path = cesiumExtensionLocation / fmt::format("cesium-trace-{}.json", timestamp);
    CESIUM_TRACE_INIT(path.string());
#endif
```

Both look kind of ugly...

apps/cesium.omniverse.app.kit
```diff
@@ -17,7 +17,7 @@ app.window.title = "Cesium for Omniverse Test App"
 app.useFabricSceneDelegate = true
 app.usdrt.scene_delegate.enableProxyCubes = false
 app.usdrt.scene_delegate.geometryStreaming.enabled = false
-app.fastShutdown = true
+app.fastShutdown = false
```
weegeekps, line 20: Can we create a second app file that derives from this one that disables `app.fastShutdown`? Perhaps something called `cesium.omniverse.app.trace.kit`? Without this option debugging some of the Python stuff can be quite a poor experience.

src/core/src/Context.cpp
weegeekps: I agree both are ugly, but your alternative option is more readable to me. Mind swapping to it?

lilleyse: I landed on something more compact:

```c++
#ifdef CESIUM_TRACING_ENABLED
    const auto timeNow = std::chrono::time_point_cast<std::chrono::microseconds>(std::chrono::steady_clock::now());
    const auto timeSinceEpoch = timeNow.time_since_epoch().count();
    const auto path = cesiumExtensionLocation / fmt::format("cesium-trace-{}.json", timeSinceEpoch);
    CESIUM_TRACE_INIT(path.string());
#endif
```

apps/cesium.omniverse.app.kit
lilleyse: Good idea, done!

PR #313, opened 2023-06-01
Avoid unnecessary allocations when adding geometry data to Fabric

lilleyse:
Instead of reading positions, texcoords, indices, etc into `pxr::VtArray`'s and then copying into Fabric arrays, we can skip the middle step. This saves about 0.1 to 0.2 ms per `FabricPrepareRenderResources::setFabricMeshes` call.

To reproduce: build with [tracing enabled](https://github.com/CesiumGS/cesium-omniverse/blob/main/docs/developer-setup/README.md#tracing), fly around the scene a bit, drag the result into https://ui.perfetto.dev/, and run the SQL query:
```
SELECT AVG(dur) / 1000000 FROM (SELECT dur FROM slice WHERE name = 'FabricPrepareRenderResources::setFabricMeshes')
```

### DIFF
--- a/src/core/include/cesium/omniverse/GltfAccessors.h
+++ b/src/core/include/cesium/omniverse/GltfAccessors.h
@@ -1,0 +1,103 @@
+#pragma once
+
+#ifdef CESIUM_OMNI_MSVC
+#pragma push_macro("OPAQUE")
+#undef OPAQUE
+#endif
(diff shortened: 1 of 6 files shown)


corybarr (approved):
Makes sense. Very clean. I can't find anything to note.

PR #314, opened 2023-06-01
Improve `launch.json` and python debugging workflow

lilleyse:
* Committed Windows and Linux versions of `launch.json` to the repo and removed them from the developer setup guide
* Updated `launch.json` to include a `Python Debugging (start)` configuration to be paired with the existing `Python Debugging (attach)`. This runs `cesium.omniverse.app.python.debug.kit` which enables "omni.kit.debug.vscode" and "omni.kit.debug.python".
* Added some more debugging tips to the developer setup guide

@corybarr can you review?


.vscode/launch.json.windows
```diff
@@ -0,0 +1,74 @@
...
+        "${workspaceFolder}/apps/cesium.omniverse.app.trace.kit"
+      ],
+      "cwd": "${workspaceFolder}",
+      "type": "cppvsdbg",
+      "request": "launch",
+      "console": "internalConsole",
+      "internalConsoleOptions": "openOnSessionStart"
+    },
+    {
+      "name": "Omniverse Code",
```
corybarr, line 31: @lilleyse Good call changing from "Code." I didn't know what this was in the context of Run and Debug options.

docs/developer-setup/README.md
```diff
@@ -697,13 +498,28 @@ Each workspace contains recommended extensions and settings for VSCode developme
...
 To run a task:
 
 - `Ctrl + Shift + B` and select the task, e.g. `Build`
 - Select the build type and compiler (if applicable)
 
+### Debugging
+
+Windows and Linux versions of `launch.json` are provided in the `.vscode` folder.
+
+* On Windows copy `launch.json.windows` and rename it to `launch.json`.
```
corybarr, line 512: The symbolic link worked here for me. The user wouldn't have to re-copy if we edit a launch.json.*. It's a bit more complex, but the user already has to do the exact same thing for the python3.exe->python.exe symlink.

corybarr (approved):

PR #315, opened 2023-06-06
Skip link_app step in build process

lilleyse:
@mattelser was running into an issue where the build process stalled at the link_app step. This is the step where the nvidia toolchain tries to find an omniverse application like Code or Create and create a symlink to it.

This used to be important for Python IntelliSense search paths, but we can point to the `kit-sdk` folder instead. The only other place that uses the symlink is the "Omniverse Code" configuration in `launch.json`, but I find myself rarely using that.


lilleyse:
@mattelser could you take a look at this?

corybarr (approved):

mattelser:
FTR there was a failure on the centos build step, the container ran out of space. This does not appear related to the change and, despite recurring when that test was rerun, identical code passed the same test. LGTM 👍🏻

PR #318, opened 2023-06-07
Update Visual Studio settings

lilleyse:
I'm about to start profiling with Visual Studio and did a few things to improve the build process. These changes are mainly relevant if you open the `cesium-omniverse` folder instead of a solution file.

* Renamed build folders to `build-vs-*` to deconflict with VSCode build folders. It's not really possible for VSCode and Visual Studio to use the same build folder since they use different generators, `Ninja Multi-Config` and `Ninja` respectively.
* Added `-v install` so that the install target is always run during build
* Added `launch.vs.json` so that we can launch and debug kit applications from within Visual Studio

I couldn't figure out how to invoke a pre-build task from `launch.vs.json`. What I wanted to do was force a rebuild whenever `Kit App` is run. If I figure that out I can open a separate PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,7 +37,8 @@ trace.csv
 .idea
 
 # Visual Studio Cache/Options
-.vs
+.vs/*
+!.vs/launch.vs.json
 
 # Installed python bindings
 exts/cesium.omniverse/cesium/omniverse/bindings/CesiumOmniversePythonBindings.cpython-37m-x86_64-linux-gnu.so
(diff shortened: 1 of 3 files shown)


corybarr:
Why the swap from the Visual Studio generator to Ninja?

corybarr (approved):

lilleyse:
> Why the swap from the Visual Studio generator to Ninja?

Ultimately it doesn't really matter, they both should work fine. Just wanted to be more consistent with the VSCode build. Also I found out that Ninja is actually the default.

weegeekps (approved):
This looks fine to me. Word of warning I didn't reboot to Windows to make sure it runs correctly.

Issue #306, opened 2023-05-31
Implement autoupdates for Curl certificate


weegeekps:
Certificate stores can inconsistently vary from system to system, especially on Linux machines. To remedy this we bundle a copy of the Mozilla CA certificate store with our application. This is currently manually updated but we need to set up a process which will automatically update it as part of the production build artifact generation.

We can get an up to date copy of the Mozilla certificate from: https://curl.se/docs/caextract.html

mattelser:
fixed with #319

PR #319, opened 2023-06-07
remove static certs file and automatically fetch certs on build

mattelser:
Addresses #306 

This calls a simple python script from the top level CMakeLists.txt. The python script fetches the certs from the [link @weegeekps provided](https://curl.se/docs/caextract.html) and places them in the same location as the old static file. 
I have also removed the cert file from the repo and added it to the `.gitignore`.

Here are some questions that seem worth asking and my answers to them. I am happy to discuss any alternative opinions/suggestions. 

### Why python instead of just CMake calling `curl`? 
`curl` is not universally installed by default, but there is already a python3 dependency (which includes `requests` by default). Also there seemed to be precedent for python scripts called from CMake lists (e.x. `copy_to_exts.py`).

### Why remove `cacert.pem` from git instead of just overwriting?
Having a default could be nice, but it generally seems cleaner to have anything auto-generated at build time not be included in version control. Plus this way there won't be random extra cert files being added to future commits any time they're auto-updated.


weegeekps (changes requested):
Looks fine except that the python file fails `flake8` formatting rules. Run `black` on it and that should fix it.

We probably need to add this folder to the flake8 configuration.

weegeekps (approved):
:rocket:

PR #320, opened 2023-06-07
Switch base material to `gltf/pbr.mdl`

lilleyse:
This PR switches our base material from `OmniPBR` to `gltf/pbr.mdl`. This is a more natural fit since all our data is coming in as glTF. Some of the advantages are:

* Texture wrapping and texcoord transforms are built into to the `base_color_texture` node which simplifies our node graph
* Alpha cutoff, alpha mode, emissive factor, and other glTF material properties are supported one-to-one

I was hoping this would be a performance win since there's less Fabric prims to create and likely a smaller shader to compile, but I didn't notice a huge difference in the tracer results or in the FPS. That said, I'm trying to set up Visual Studio to do some real profiling.

Not supported in this PR but made possible by this PR:

* Vertex colors
* Normal textures
* Occlusion textures
* Emissive textures
* PBR Next


src/core/src/FabricGeometry.cpp
```diff
@@ -282,7 +288,8 @@ void FabricGeometry::setTile(
     *worldOrientationFabric = worldOrientation;
     *worldScaleFabric = worldScale;
 
-    displayColorFabric[0] = hasMaterial ? MATERIAL_LOADING_COLOR : DEFAULT_COLOR;
+    displayColorFabric[0] = DEFAULT_VERTEX_COLOR;
```
lilleyse, line 291: `gltf/pbr.mdl` treats `displayColor` as the default vertex color so I had to change this. Otherwise the color was being multiplied by `(0, 0, 0)` and was being rendered black.

Ideally we would set it to black initially while the material is loading, then white once it's finished, but we can't do that since there's no callback when the material finishes.

weegeekps (approved):
Looks great except that the default lighting setup is a bit washed out looking. That's not really in our control and I think an intensity of 3000 is probably a bit much period, but something to be aware of.

Issue #235, opened 2023-03-21
Globe anchoring


lilleyse:
Add globe anchoring similar to Unreal and Unity. See https://cesium.com/learn/unity/unity-placing-objects/

We should use cesium-native's `CesiumGlobeAnchor` as a starting point. But first we need to update cesium-native: https://github.com/CesiumGS/cesium-omniverse/issues/234

weegeekps:
@lilleyse Let me know if you want me to break this up into smaller chunks, but I'm getting very close to done here and I'm not sure there's much to break up at this point.

PR #322, opened 2023-06-07
Broke out georeference attributes from Cesium prim into dedicated georeference prim

weegeekps:
As part of the effort for implementing #235, I realized today that I need the georeference information to be broken out of the Cesium prim into it's own dedicated prim that we can point a relationship to. A fast-follow PR will move the tilesets to also use a relationship binding to choose which georeference prim we use for that tileset.

This is also our first breaking change. Since we are still a pre-release, we are not gracefully upgrading old USDs files, so users will need to beware and we'll need to call this out when we release. We also need to update the samples project once this is merged. I can do so once this is merged.

### DIFF
--- a/exts/cesium.omniverse/cesium/omniverse/ui/attributes/__init__.py
+++ b/exts/cesium.omniverse/cesium/omniverse/ui/attributes/__init__.py
@@ -1,3 +1,4 @@
 from .data_attributes_widget import CesiumDataSchemaAttributesWidget  # noqa: F401
+from .georeference_attributes_widget import CesiumGeoreferenceSchemaAttributesWidget  # noqa :F401
 from .tileset_attributes_widget import CesiumTilesetAttributesWidget  # noqa: F401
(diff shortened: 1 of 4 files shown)


lilleyse (approved):
Looks great. The tileset UI doesn't show the `georeferenceBinding` property but I assume that will be added in the follow up PR when the functionality is added.

weegeekps:
> Looks great. The tileset UI doesn't show the `georeferenceBinding` property but I assume that will be added in the follow up PR when the functionality is added.

Yeah, the tileset doesn't actually use any binding yet. We're basically doing the same thing, but I hope to have that other PR open today.

For V1 of this, we're going to bind to `/World/CesiumGeoreference` but when we support multiple georeferences later, we can fix this and expose a UI method for binding.

weegeekps:
Oof, I just merged without waiting for two reviews, per our earlier discussion. Sorry.

PR #329, opened 2023-06-09
Ensure requests exists

### LINKED ISSUE #327: ensure python requests is installed
#306 introduced a dependency on the `requests` module in order to have certs installed, but `requests` is only installed by default on linux. It's silently failing during the windows build process. I am not yet sure if this causes a fallback to the default certs on the machine or causes failures, but either way we should ensure the requests module is available.

mattelser:
fixes #327. `update_certs.py` was silently failing on windows


mattelser:
closing so I can squash these tiny commits and resubmit a PR

PR #331, opened 2023-06-09
update_certs.py was silently failing on windows

### LINKED ISSUE #327: ensure python requests is installed

mattelser:
fixes #327. It needed `requests` and to explicitly write the cert file as `utf-8`

### DIFF
--- a/scripts/update_certs.py
+++ b/scripts/update_certs.py
@@ -33,7 +33,8 @@ def main():
         print(f"failed to fetch certificates from {CERT_URL}")
         return -1
 
-    with open(CERT_FILE_PATH, "w") as f:
+    # explicit encoding is required for windows
+    with open(CERT_FILE_PATH, "w", encoding='utf-8') as f:
         f.write(req.text)
 
     return 0
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -115,12 +115,13 @@ jobs:
       # hardcoding their locations instead. Followed these instructions for adding to the PATH on PowerShell:
       # https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#environment-files
       # Note that this is only a problem on CI, not with local builds
+      # Also requests is not installed by default on windows, but is expected by scripts/update_certs.py
       - name: Install Windows dependencies
         if: runner.os == 'Windows'
         run: |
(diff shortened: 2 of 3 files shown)


lilleyse:
Thanks @mattelser.

Could you also update the Windows installation steps in the developer setup README? Both the "Install manually" and "Install with Chocolatey" sections.

scripts/update_certs.py
```diff
@@ -33,7 +33,8 @@ def main():
         print(f"failed to fetch certificates from {CERT_URL}")
         return -1
 
-    with open(CERT_FILE_PATH, "w") as f:
+    # explicit encoding is required for windows
+    with open(CERT_FILE_PATH, "w", encoding='utf-8') as f:
```
lilleyse, line 37: Does this fix this error?

```
1> [CMake] Traceback (most recent call last):
1> [CMake]   File "C:\Users\slilley\Code\cesium-omniverse\scripts\update_certs.py", line 43, in <module>
1> [CMake]     sys.exit(main())
1> [CMake]              ^^^^^^
1> [CMake]   File "C:\Users\slilley\Code\cesium-omniverse\scripts\update_certs.py", line 37, in main
1> [CMake]     f.write(req.text)
1> [CMake]   File "C:\Python311\Lib\encodings\cp1252.py", line 19, in encode
1> [CMake]     return codecs.charmap_encode(input,self.errors,encoding_table)[0]
1> [CMake]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
1> [CMake] UnicodeEncodeError: 'charmap' codec can't encode character '\u0151' in position 37587: character maps to <undefined>
```

mattelser: Yes

mattelser:
> Could you also update the Windows installation steps in the developer setup README? Both the "Install manually" and "Install with Chocolatey" sections.

good call, fixed

lilleyse (approved):
Confirmed that this is working on windows after doing a `git clean -dxf`

PR #337, opened 2023-06-12
Tileset georeference binding

weegeekps:
Another step towards getting #235 implemented.

Tilesets now have the Georeference prim they're using property related, and this relation is used during most of the rendering step. It's important to note this does not mean we support multiple-georeference points yet. I had concerns about removing the direct call to the Georeference prim in `Context::onUpdateFrame` for performance reasons, so I didn't change that. Additionally, on the reload step we are referencing the Georeference prim directly there for potential performance reasons as well.

### DIFF
--- a/src/core/src/Context.cpp
+++ b/src/core/src/Context.cpp
@@ -215,7 +215,7 @@ void Context::reloadStage() {
     for (const auto& prim : stage->Traverse()) {
         const auto& path = prim.GetPath();
         if (UsdUtil::isCesiumTileset(path)) {
-            AssetRegistry::getInstance().addTileset(path, pxr::SdfPath("/CesiumGeoreference"));
+            AssetRegistry::getInstance().addTileset(path, UsdUtil::GEOREFERENCE_PATH);
         } else if (UsdUtil::isCesiumImagery(path)) {
             AssetRegistry::getInstance().addImagery(path);
         }
--- a/src/core/src/Context.cpp
+++ b/src/core/src/Context.cpp
@@ -214,7 +214,7 @@ void Context::reloadStage() {
     for (const auto& prim : stage->Traverse()) {
         const auto& path = prim.GetPath();
         if (UsdUtil::isCesiumTileset(path)) {
-            AssetRegistry::getInstance().addTileset(path);
+            AssetRegistry::getInstance().addTileset(path, pxr::SdfPath("/CesiumGeoreference"));
         } else if (UsdUtil::isCesiumImagery(path)) {
             AssetRegistry::getInstance().addImagery(path);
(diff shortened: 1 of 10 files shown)


src/core/src/Context.cpp
```diff
@@ -214,7 +215,7 @@ void Context::reloadStage() {
     for (const auto& prim : stage->Traverse()) {
         const auto& path = prim.GetPath();
         if (UsdUtil::isCesiumTileset(path)) {
-            AssetRegistry::getInstance().addTileset(path);
+            AssetRegistry::getInstance().addTileset(path, pxr::SdfPath("/CesiumGeoreference"));
```
lilleyse, line 218: It would be nice not to have to hardcode `"/CesiumGeoreference"` in multiple places. Would something like this work?

```c++
const auto georeference = UsdUtil::getOrCreateCesiumGeoreference();
AssetRegistry::getInstance().addTileset(path, georeference.GetPath());
```

src/core/src/UsdUtil.cpp
```diff
@@ -370,8 +370,10 @@ pxr::CesiumSession getOrCreateCesiumSession() {
 }
 
 pxr::CesiumGeoreference getOrCreateCesiumGeoreference() {
-    static const auto path = pxr::SdfPath("/CesiumGeoreference");
+    return getCesiumGeoreference(pxr::SdfPath("/CesiumGeoreference"));
+}
 
+pxr::CesiumGeoreference getCesiumGeoreference(const pxr::SdfPath& path) {
```
lilleyse, line 376: The other `getCesiumXXXX` functions assert if the prim doesn't exist or is invalid. Should this behave the same way? The creation code would move into `getOrCreateCesiumGeoreference`.

lilleyse (commented):
This looks great. Just minor comments from me.

src/core/src/Context.cpp
weegeekps: I was trying to go for the most performant route possible here. Ultimately, using `getOrCreateCesiumGeoreference` is going to require getting the stage again and getting the `CesiumGeoreference` object before we get what we need, which in this case is just the path.

It's worth pointing out, all of this for reload stage has to change in order for us to support multiple georeference origins, so I was leaning towards what will help us now rather than what will help us in the future.

I do agree with you that this shouldn't be hardcoded in multiple places though, and I've put the value in a static constant in `UsdUtil`.

lilleyse: The static constant is fine, though I wouldn't be too concerned about performance in `reloadStage` since it's called infrequently.

lilleyse (approved):

PR #338, opened 2023-06-13
Add vertex color support

lilleyse:
Now that we're using `gltf/pbr.mdl` (https://github.com/CesiumGS/cesium-omniverse/pull/320) it was pretty straightforward to add vertex color support.

This supports all the different vertex color variants in glTF with a few limitations:

* Only `COLOR_0` is supported
* Alpha values are ignored. I tried treating `vertexColor` primvar as a 4-channel color but it didn't work. The diff is here if we ever revisit this in the future: [vec4-diff.txt](https://github.com/CesiumGS/cesium-omniverse/files/11737515/vec4-diff.txt)

image

To test: extract [vertex-color.zip](https://github.com/CesiumGS/cesium-omniverse/files/11737425/vertex-color.zip), edit the url in the .usda file for your system, and drag the .usda onto the stage.

vertex-colors-batched


src/core/src/GltfUtil.cpp
```diff
@@ -253,6 +253,59 @@ TexcoordsAccessor getTexcoords(
...
+            if (view.status() == CesiumGltf::AccessorViewStatus::Valid) {
+                return VertexColorsAccessor(view);
+            }
+        } else if (vertexColorAccessor->type == CesiumGltf::AccessorSpec::Type::VEC4) {
+            CesiumGltf::AccessorView<glm::fvec4> view{model, *vertexColorAccessor};
+            if (view.status() == CesiumGltf::AccessorViewStatus::Valid) {
+                return VertexColorsAccessor(view);
+            }
+        }
+    }
```
weegeekps, line 304: NIT: There is a lot of repetition here, and it feels like we should be able to do something to cut down on it a bit.

weegeekps (approved):
:ship:

PR #341, opened 2023-06-14
Created Cesium Power Tools extension with a sample tool

weegeekps:
This creates the discussed "Cesium Power Tools" extension for experimental helper tools that we've been discussing. I've created a single sample tool which simply opens the Cesium Debugger window.

The main window for the Power Tools is the `CesiumPowertoolsWindow` class, and this is designed to just basically be a simple way to access tools themselves by providing an easy way to create buttons for those tools. Your tool generally should not be added directly to the `CesiumPowertoolsWindow` class, and instead should be within another window or class. Think of `CesiumPowertoolsWindow` as a toolbar of sorts.

Several approaches can be used to add a new tool, but the one I had in mind when I set this up is:

1. You create a new window class for your tool.
2. The window class contains a static method for creating an instance of said window. We do not hold on to that reference anywhere as we shouldn't need it.
3. You create a new `PowertoolsAction` in the list in `CesiumPowertoolsWindow` which calls the static method for opening the window.
4. You implement your tool within that window.

Ultimately, I do expect this pattern to change a bit as we learn more and create tools, but this process should at least get us started and allow the tools to be independent of one another.

As part of this PR, I did make some changes to the main extension, moving `CesiumOmniverseExtension._dock_window_async` into the utilities, and adding proper imports in `utils/_init_.py` to ensure they can be accessed outside the `cesium.omniverse` package.


exts/cesium.powertools/doc/README.md
```diff
@@ -0,0 +1,13 @@
+# Cesium for Omniverse Power Tools
+
+This extension provides supporting features and tools that may be useful while developing apps and stages using Cesium for Omniverse, but are provided **on an entirely your milage may vary** basis. All of these tools are considered experimental. If you like a particular tool in this extension, and would like to see it moved to the main Cesium for Omniverse extension, please [feel free to let us know on the Cesium for Omniverse forums](https://community.cesium.com/c/cesium-for-omniverse/14).
+
+## Current Features
+
+### Open Cesium Debugger Button
```
lilleyse, line 7:
```suggestion
### Open Cesium Debugging Window
```

exts/cesium.powertools/cesium/powertools/powertools_window.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __init__(self, **kwargs):
+        super().__init__(CesiumPowertoolsWindow.WINDOW_NAME, **kwargs)
+
+        self._logger = logging.getLogger(__name__)
+
+        # You do not necessarily need to create an action function in this window class. If you have a function
+        #  in another window or class, you can absolutely call that instead from here.
+        self._actions: List[PowertoolsAction] = [
+            PowertoolsAction("Cesium Debugger", CesiumOmniverseDebugWindow.show_window)
```
lilleyse, line 37:
```suggestion
            PowertoolsAction("Open Cesium Debugging Window", CesiumOmniverseDebugWindow.show_window)
```

lilleyse (approved):
The "Open Cesium Debugging Window" button is already super handy for me. 😄 

Just one naming suggestion, otherwise this is good to go.

weegeekps:
@lilleyse Good suggestions, thank you.

PR #343, opened 2023-06-14
More statistics in Cesium Debugging window

lilleyse:
Adds more statistics to help with profiling:

* For geometries and materials, distinguish between capacity (pool size) vs. loaded. Previously only the capacity was shown.
  * Note that we aren't using a geometry pool in Kit 104.2, which is why the numbers are equal
* Renamed "visible" to "rendered"
* Adds more stats that you would find in Cesium for Unity and Cesium for Unreal
  * Tiles visited
  * Culled tiles visited
  * Tiles rendered
  * Tiles culled
  * Max depth visited
  * Tiles loading (worker)
  * Tiles loading (main)
  * Tiles loaded

image


weegeekps (approved):
LGTM

Issue #261, opened 2023-04-25
Performance testing application


lilleyse:
As we look at https://github.com/CesiumGS/cesium-omniverse/pull/252 and mesh/material pooling it would be helpful to have a performance testing application that lets us measure performance in a repeatable way.

To start the application might 

* Load a stage with Cesium World Terrain
* Fly the camera from point A to B
* Record the average FPS, number of tiles loaded, and other useful stats

It would likely be a Kit application similar to `cesium.omniverse.app.kit` that depends on `cesium.omniverse` and runs some Python code. Depending on what it needs to do with `cesium.omniverse` we might need to start exposing a Python API: https://github.com/CesiumGS/cesium-omniverse/issues/12.

PR #347, opened 2023-06-15
Basic project structure for performance testing app

lilleyse:
Part of #261 

Adds the basic project structure for a performance testing app. I'm looking for feedback on this before I start writing the actual performance testing code.

I decided to extend `cesium.omniverse.app.kit` rather than create a completely new app. In the future I might explore `omni.kit.quicklayout` and creating a custom UI, but for now I think it's best to keep it simple.

The Python code creates a "Cesium Performance Testing" window similar to `cesium.powertools`.

One thing I'm not sure about is whether the `cesium.performance` extension should live in the `cesium-omniverse/app` folder like it is now, or if it should live in the `cesium-omniverse/exts` folder with the other extensions. Any thoughts on that @weegeekps?


apps/exts/cesium.performance/cesium/performance/performance_window.py
```diff
@@ -0,0 +1,20 @@
...
+        super().__init__(CesiumPerformanceWindow.WINDOW_NAME, **kwargs)
+
+        self._logger = logging.getLogger(__name__)
+
+        self.frame.set_build_fn(self._build_fn)
+
+    def destroy(self) -> None:
+        super().destroy()
+
+    def _build_fn(self):
```
weegeekps, line 19: NIT: This can be marked static.

apps/exts/cesium.performance/config/extension.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+title = "Cesium for Omniverse Performance Testing Extension"
+version = "0.0.0"
+app = false
```
weegeekps, line 4: We should add `toggleable = false`.

lilleyse: Is there a reason for setting `toggleable = false`? I think it's fine if someone decides to disable the extension.

For example, `omni.code.app.setup` doesn't set `toggleable = false` either.

apps/exts/cesium.performance/cesium/performance/performance_window.py
lilleyse: Ah, so this function is a stub right now but in the next PR I'm actually accessing stuff from `self` to build out the UI.

lilleyse:
> One thing I'm not sure about is whether the `cesium.performance` extension should live in the `cesium-omniverse/app` folder like it is now, or if it should live in the `cesium-omniverse/exts` folder with the other extensions.

@weegeekps thoughts on this question? I can't decide which approach I prefer.

weegeekps:
@lilleyse I like it living in the apps folder. Prevents any confusion about what it's for.

Thinking about it though, you may want to rename the module `cesium.omniverse.app.performance` just to further clarify that it's an app extension.

lilleyse:
Good idea. I ended up renaming the app to `cesium.performance.kit` and the extension to `cesium.performance.app`. This seems to be consistent with the Code and Create naming convention (see below).

I tried renaming the app to `cesium.omniverse.performance.kit` and the extension to `cesium.omniverse.performance.app` but it couldn't load due to module conflicts with the `cesium.omniverse` extension. I'm not sure if prefixing Python-based apps with `cesium.omniverse` will work.

I also think we should remove the word `app` from the other `.kit` files, e.g. rename `cesium.omniverse.app.kit` to `cesium.omniverse.dev.kit` or something, but I didn't want to do it in this PR.

#
### App
`(\w+)\.(\w+)\.kit`

E.g.

- cesium.performance.kit

### App extension
`(\w+)\.(\w+)\.app(\.(\w+))?`

E.g. 

- cesium.performance.app

... or if there are sub extensions

- cesium.performance.app.setup
- cesium.performance.app.resources

apps/exts/cesium.performance/config/extension.toml
weegeekps: I guess there's no harm in it. My fear is that if a user disables it and then reenables it, the startup stuff will execute again, but that may be harmless. Seems undesirable regardless.

lilleyse: Alright, I added `toggleable = false`. I don't see much harm either way.

weegeekps (approved):
:shipit:

lilleyse:
@weegeekps Follow up PR here: https://github.com/CesiumGS/cesium-omniverse/pull/357

PR #349, opened 2023-06-16
Add Georeference Helper powertool

r-veenstra:
Currently if a user has CAD data in a projected coordinate system, they need to work out what lat/long/height aligns to their projects origin.  This is almost always a manual calculation that involves visiting a site like EPSG.io, copy/pasting various coordinates in, then copy/pasting lat/longs into the Cesium Georeference prim (usually resulting in an error because of human error copying the a value into the wrong field) 

This tool is designed to replace the need for this and allow the user to convert coords directly in Omniverse using the proj library.  It also lets them put the calculated values straight into Cesium Georeference to minimise human error.  Potentially some of this logic could eventually make its way into Globe Anchors, where a user could specified the EPSG for the globe anchor coordinates in a similar way to this tool.

This is also an experiment in working with the new power tools extension, so I'm particularly interested in feedback around general structure, how we organise code and UI if we have a dozen similar tools, etc.

Some notes on implementation and things to do

- I've followed a similar folder structure to the main Cesium Omniverse extension - ui folder, utils folder, etc.  Happy to adjust this if needed
- I'm not 100% decided on how we handle managing windows yet.  Currently the Power tools extension just creates a new one when clicked, the window itself isn't added to the Omniverse menu or docked to the UI.  Its currently not tracked or cleaned up by the power tools destroy() function.  TBD
- This tool relies on pyproj being installed by pip.  I had some trouble using omni.kit.pipapi.install(), I have since changed to declaring the requirement in extension.toml.  Would be useful for someone to test on their end to see if this successfully installs pyproj
- A host of UX things like remembering the values between window loads, but for now the experience is no worse that visiting coordinate calculator in your browser, so I'm okay with this
- Still need to implement a workflow to specify if the users project data is offset from the Geoid or Mean Sea Level
- Probably also add a geocoding search so users can find a lat/long by searching Bing


exts/cesium.powertools/cesium/powertools/powertools_window.py
```diff
@@ -50,3 +51,6 @@ def _build_fn(self):
         with ui.VStack(spacing=4):
             for action in self._actions:
                 action.button()
+
+    def _open_georeference_helper(self):
+        CesiumGeorefHelperWindow(width=250, height=550)
```
weegeekps, line 56: Move this to a public static method on `CesiumGeorefHelperWindow`.

exts/cesium.powertools/cesium/powertools/ui/georef_helper_window.py
```diff
@@ -0,0 +1,149 @@
...
+        # Get the CRS and check if it is valid, adjust UI values accordingly
+        crs = get_crs_name_from_epsg(self._epsg_model.get_value_as_int())
+        if crs is None:
+            self._epsg_name_model.set_value("Invalid EPSG Code")
+            self._wgs84_latitude_model.set_value(math.nan)
+            self._wgs84_longitude_model.set_value(math.nan)
+            self._wgs84_height_model.set_value(math.nan)
+            self._ecef_x_model.set_value(math.nan)
+            self._ecef_y_model.set_value(math.nan)
+            self._ecef_z_model.set_value(math.nan)
```
weegeekps, line 39: I think this is maybe the second or third time I've actually seen someone use `math.nan` correctly, so bravo.

exts/cesium.powertools/cesium/powertools/ui/georef_helper_window.py
```diff
@@ -0,0 +1,149 @@
...
+            self._wgs84_latitude_model = ui.SimpleFloatModel(0.0)
+            self._wgs84_longitude_model = ui.SimpleFloatModel(0.0)
+            self._wgs84_height_model = ui.SimpleFloatModel(0.0)
+            float_field_with_label("Latitude", model=self._wgs84_latitude_model, enabled=False)
+            float_field_with_label("Longitude", model=self._wgs84_longitude_model, enabled=False)
+            float_field_with_label("Elevation", model=self._wgs84_height_model, enabled=False)
+
+            ui.Spacer(height=10)
+
+            ui.Label("ECEF Results:", height=20)
```
weegeekps, line 138: I'd suggest setting `height=0` on these and instead using a style that contains increases the font size and applies bold on it.

weegeekps (commented):
I like this, but in addition to my comments I have some thoughts on the layout. I don't think we should be using the `*.ui`, `*.util`, etc. pattern here. Instead I'd argue that we try to keep each powertool as separate as possible.

So, for this tool, you'd have a package like `cesium.powertools.georefhelper` and then under that you can include the files you need to make that work. In the event that we decide to "promote" a tool to the main extension, this will make it much easier to do so as we won't have to sort through the code to determine what we need to take and what we don't.

exts/cesium.powertools/config/extension.toml
```diff
@@ -27,3 +27,8 @@ icon = "doc/resources/icon.png"
 
 [[python.module]]
 name = "cesium.powertools"
+
+[python.pipapi]
+requirements = [
+    "pyproj==3.6.0"
+]
```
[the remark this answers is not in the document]
lilleyse: @corybarr since you're also on Windows, could you check if you get this same error?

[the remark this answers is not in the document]
corybarr: @lilleyse Confirmed: 
image

exts/cesium.powertools/cesium/powertools/ui/georef_helper_window.py
r-veenstra: Done - though I couldn't see a bold specific style setting.  There is an option to specify different TTF's, wasn't sure if there were fonts built in or I needed to supply, in the end for the sake of time I just increased the heading font size slightly which works ok.

exts/cesium.powertools/cesium/powertools/powertools_window.py
r-veenstra: Done - new method is called create_window

exts/cesium.powertools/config/extension.toml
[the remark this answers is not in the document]
r-veenstra: This is resolved.  The .toml declaration might be a kit 105 thing, either way it wasn't working in my 104 setup.  I reverted back to calling omni.kit.pipapi.install(), tested by removing my local dependencies and it installs ok for me when enabling the extension.

weegeekps (approved):
LGTM

PR #350, opened 2023-06-16
Add `disableTextures` option to `/Cesium` prim

lilleyse:
For performance testing it's useful to measure differences between:

* No materials
* Materials without textures
* Materials with textures

To help with that, this PR adds a new option called `cesium:debug:disableTextures` to the `/Cesium` prim.


weegeekps (approved):
LGTM

PR #351, opened 2023-06-16
Use Cesium USD python bindings instead of generic USD functions

lilleyse:
For the performance testing app I'm doing a bunch of manipulation of the USD stage and have tried to use the Cesium USD python bindings instead of the generic USD functions.

In `cesium.omniverse` I noticed one area that could be switched to use the Cesium bindings.

I also added `cesium.usd.plugins` to the Pylance search paths for VSCode. @corybarr this should fix the IntelliSense errors you were seeing. Though it's limited because we don't have complete Python stubs yet (see [this TODO](https://github.com/CesiumGS/cesium-omniverse/blob/bcd70e23dd4052927a3911a9ab1d7cdc9a84aa34/exts/cesium.usd.plugins/cesium/usd/plugins/CesiumUsdSchemas/__init__.pyi#L12)).

### DIFF
--- a/.vscode/cesium-omniverse-linux.code-workspace
+++ b/.vscode/cesium-omniverse-linux.code-workspace
@@ -31,6 +31,7 @@
     // Python modules search paths:
     "python.analysis.extraPaths": [
       "${workspaceFolder}/exts/cesium.omniverse",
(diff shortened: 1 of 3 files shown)


corybarr (approved):
Looks great and will be handy. It's nice seeing the string arg get replaced with a method call.

PR #352, opened 2023-06-16
Create CesiumTileset concrete prim type

corybarr:
This changes the Type of Cesium tileset prims from an Xform to CesiumTilesetPrim:

image

The new prim inherits from UsdGeomBoundable, which allows us to use properties such as `extent `to better interact with the viewport.


exts/cesium.omniverse/cesium/omniverse/ui/attributes/tileset_attributes_widget.py
```diff
@@ -3,12 +3,12 @@
 from omni.kit.property.usd.usd_property_widget import SchemaPropertiesWidget
 import omni.ui as ui
 from ...bindings import ICesiumOmniverseInterface
-from cesium.usd.plugins.CesiumUsdSchemas import TilesetAPI as CesiumTilesetAPI
+from cesium.usd.plugins.CesiumUsdSchemas import Tileset as CesiumTilesetPrim
```
lilleyse, line 6: The `Prim` suffix can be removed - we don't include it in the Python code

```suggestion
from cesium.usd.plugins.CesiumUsdSchemas import Tileset as CesiumTileset
```

exts/cesium.omniverse/cesium/omniverse/ui/attributes/tileset_attributes_widget.py
```diff
@@ -3,12 +3,12 @@
...
 
 
 class CesiumTilesetAttributesWidget(SchemaPropertiesWidget):
     def __init__(self, _cesium_omniverse_interface: ICesiumOmniverseInterface):
-        super().__init__("Cesium Tileset Settings", CesiumTilesetAPI, include_inherited=False)
+        super().__init__("Cesium Tileset Settings", CesiumTilesetPrim, include_inherited=False)
```
lilleyse, line 11:
```suggestion
        super().__init__("Cesium Tileset Settings", CesiumTileset, include_inherited=False)
```

src/core/src/UsdUtil.cpp
```diff
@@ -294,32 +294,11 @@ pxr::CesiumGeoreference defineCesiumGeoreference(const pxr::SdfPath& path) {
...
-    tileset.CreateMaximumSimultaneousTileLoadsAttr(pxr::VtValue(uint32_t(20)));
-    tileset.CreateMaximumCachedBytesAttr(pxr::VtValue(uint64_t(536870912)));
-    tileset.CreateLoadingDescendantLimitAttr(pxr::VtValue(uint32_t(20)));
-    tileset.CreateEnableFrustumCullingAttr(pxr::VtValue(true));
-    tileset.CreateEnableFogCullingAttr(pxr::VtValue(true));
-    tileset.CreateEnforceCulledScreenSpaceErrorAttr(pxr::VtValue(true));
-    tileset.CreateCulledScreenSpaceErrorAttr(pxr::VtValue(64.0f));
-    tileset.CreateSuspendUpdateAttr(pxr::VtValue(false));
-    tileset.CreateSmoothNormalsAttr(pxr::VtValue(false));
-    tileset.CreateShowCreditsOnScreenAttr(pxr::VtValue(false));
```
lilleyse, line 321: Related to the discussion in https://github.com/CesiumGS/cesium-omniverse/pull/339#discussion_r1228761878...

I think these lines should be added back unless they're a reason to remove them.

src/core/src/UsdUtil.cpp
```diff
@@ -450,7 +429,8 @@ bool isCesiumTileset(const pxr::SdfPath& path) {
         return false;
     }
 
-    return prim.HasAPI<pxr::CesiumTilesetAPI>();
+    // return prim.HasAPI<pxr::CesiumTileset>();
+    return true;
```
lilleyse, line 433:
```suggestion
    return prim.IsA<pxr::CesiumTileset>();
```

exts/cesium.usd.plugins/cesium/usd/plugins/CesiumUsdSchemas/__init__.pyi
```diff
@@ -5,22 +5,27 @@ __all__ = [
...
     pass
 
+
 class Imagery(pxr.Usd.Typed, pxr.Usd.SchemaBase, Boost.Python.instance):
     pass
 
-class TilesetAPI(pxr.Usd.APISchemaBase, pxr.Usd.SchemaBase, Boost.Python.instance):
+
+class Tileset(pxr.Usd.Typed, pxr.Usd.SchemaBase, Boost.Python.instance):
     pass
```
lilleyse, line 27: Could you also add `Session` to this file? I forgot to do that in https://github.com/CesiumGS/cesium-omniverse/pull/295

exts/cesium.usd.plugins/schemas/cesium_schemas.usda
```diff
@@ -129,12 +130,12 @@ class CesiumGeoreferencePrim "CesiumGeoreferencePrim" (
     )
 }
 
-class "CesiumTilesetSchemaAPI" (
+class CesiumTilesetPrim "CesiumTilesetPrim" (
     doc = """Adds Cesium specific data to a prim for representing a tileset."""
```
lilleyse, line 134: Now that it's not an API schema the description can be simplified:

```suggestion
    doc = """A prim representing a tileset."""
```

src/plugins/CesiumUsdSchemas/src/CesiumUsdSchemas/wrapTileset.cpp
```diff
@@ -27,148 +27,148 @@ WRAP_CUSTOM;
...
-void wrapCesiumTilesetAPI()
+void wrapCesiumTileset()
 {
-    typedef CesiumTilesetAPI This;
+    typedef CesiumTileset This;
 
-    class_<This, bases<UsdAPISchemaBase> >
-        cls("TilesetAPI");
+    class_<This, bases<UsdTyped> >
+        cls("Tileset");
```
lilleyse, line 171: You may need to rerun `regenerate_schema.bat`. It looks like `tokens.h` and `wrapTileset.cpp` have a few more changes after I run it.

lilleyse (commented):
This looks good @corybarr - I'm looking forward to getting this and then https://github.com/CesiumGS/cesium-omniverse/pull/339 merged. A few comments from me, but nothing major.

src/core/src/UsdUtil.cpp
corybarr: I don't think the lines above have any effect, except for the possibility of altering default values. These attributes are created and default values set in the schema file. For example, if we look at [MaximumScreenSpaceError](https://github.com/CesiumGS/cesium-omniverse/pull/352#discussion_r1234055441), then to keep the code in sync we'd need to make sure the default in the schema matches the C++. If the C++ differs, then the UI will show the default value has changed and enable to button to reset the value.

exts/cesium.usd.plugins/cesium/usd/plugins/CesiumUsdSchemas/__init__.pyi
corybarr: Ah, apologies. I missed the note on the TODO I wrote to add that. Added.

src/plugins/CesiumUsdSchemas/src/CesiumUsdSchemas/wrapTileset.cpp
corybarr: Good call. Fixed.

src/core/src/UsdUtil.cpp
lilleyse: Ah, makes sense. I don't think we have any reason to set different default values in C++.

One thing I noticed is that creating attributes explicitly causes them to be saved in the .usda, whereas the other approach only saves values that are set. I would guess the second approach is the more USD way to do things.

**Create Attributes**
```
def Xform "Cesium_Tileset" (
    prepend apiSchemas = ["CesiumTilesetSchemaAPI"]
)
{
    float cesium:culledScreenSpaceError = 64
    bool cesium:enableFogCulling = 1
    bool cesium:enableFrustumCulling = 1
    bool cesium:enforceCulledScreenSpaceError = 1
    bool cesium:forbidHoles = 0
    prepend rel cesium:georeferenceBinding = </CesiumGeoreference>
    string cesium:ionAccessToken = ""
    int64 cesium:ionAssetId = 1
    uint cesium:loadingDescendantLimit = 20
    uint64 cesium:maximumCachedBytes = 536870912
    float cesium:maximumScreenSpaceError = 16
    uint cesium:maximumSimultaneousTileLoads = 20
    bool cesium:preloadAncestors = 1
    bool cesium:preloadSiblings = 1
    bool cesium:showCreditsOnScreen = 0
    bool cesium:smoothNormals = 0
    uniform token cesium:sourceType = "ion"
    bool cesium:suspendUpdate = 0
    string cesium:url
}
```

**Don't Create Attributes**
```
def Xform "Cesium_Tileset" (
    prepend apiSchemas = ["CesiumTilesetSchemaAPI"]
)
{
    prepend rel cesium:georeferenceBinding = </CesiumGeoreference>
    string cesium:ionAccessToken = ""
    int64 cesium:ionAssetId = 1
}
```

src/plugins/CesiumUsdSchemas/src/CesiumUsdSchemas/tileset.h
```diff
@@ -24,10 +24,10 @@ PXR_NAMESPACE_OPEN_SCOPE
...
 
 // -------------------------------------------------------------------------- //
-// CESIUMTILESETSCHEMAAPI                                                     //
+// CESIUMTILESETPRIM                                                          //
 // -------------------------------------------------------------------------- //
 
-/// \class CesiumTilesetAPI
+/// \class CesiumTileset
 ///
 /// Adds Cesium specific data to a prim for representing a tileset.
```
lilleyse, line 32: I re-ran `regenerate_schema.bat` and all looked good, except this doc which needed to be updated.

```suggestion
/// A prim representing a tileset.
```

lilleyse (commented):
@corybarr Just these two comments and then this is good to go from my perspective

- https://github.com/CesiumGS/cesium-omniverse/pull/352#discussion_r1235253659
- https://github.com/CesiumGS/cesium-omniverse/pull/352#discussion_r1235245652

src/core/src/UsdUtil.cpp
corybarr: Good call. We should. How about doing it as a small PR right after this? It seems like it would keep the commit log tidy.

lilleyse (approved):

src/core/src/UsdUtil.cpp
corybarr: >One thing I noticed is that creating attributes explicitly causes them to be saved in the .usda, whereas the other approach only saves values that are set. 

Interesting. Then by not creating attributes in C++ we have the possibility of altering some values that would still auto-set on old USD files, like cache size or something. Even if we go the other route and keep the calls, it's probably more correct to get and set the attributes, since they're already created. We can move this discussion to the new PR.

PR #356, opened 2023-06-20
Remove Create*Attr calls covered in USD schema

corybarr:
Some C++ code was duplicating functionality provided by the USD schema files for creating attributes and setting their defaults. The C++ calls and USD schemas were checked to make sure all attributes are still being created and defaults are the same values.


lilleyse:
For future reference, here's the discussion where this started: https://github.com/CesiumGS/cesium-omniverse/pull/352#discussion_r1234055441

lilleyse (approved):
Looks good

PR #358, opened 2023-06-20
Fix advanced install instructions

lilleyse:
Updated the instructions for installing cesium-omniverse to a standalone directory.

* Had to separate the build step from install step
* Remove `-D CMAKE_INSTALL_PREFIX` in favor of `--prefix`
* Window and Linux have slightly different instructions since Windows uses a multi-configuration generator.


mattelser (approved):
🚢

Issue #309, opened 2023-05-31
Generate Python Stubs for C++ Bindings


weegeekps:
We have a lot of Python code calling our C++ native code, but no real stubs to help with auto-completion and safety. We need to figure out how to get one of the typical tools generating the bindings *or* find another way to generate the bindings, then implement that as a script that can be run on-demand similar to the schema updates.

lilleyse:
@weegeekps would generating python stubs also fix https://github.com/CesiumGS/cesium-omniverse/issues/283?

corybarr:
@weegeekps Fixing this would be great. Aside from ongoing dev, it would remove a large slowdown when learning the codebase for the first time. Let me know if you need help.

mattelser:
I've tried several things to autogenerate stubs, but have had no luck. It's entirely possible that this can be done using one of the tools I've tried and that I'm just using them wrong, so for posterity here's what I've tried:
- `pybind11-stubgen` - works by importing a module. I couldn't get it to import any of the cesium omniverse modules, even with some `PATH`/`PYTHONPATH` manipulation (i.e. using `extern/nvidia/_build/target-deps/python`). It is worth calling out that this tool seems to be merging into `mypy`.
- `mypy stubgen` - works by importing a module or package. This was more promising, in that I got it to output stub files. The stub files only had imports and no class definitions though (and sometimes were just empty). Getting this far took a lot of finagling though. There are assuredly cleaner ways of doing this, but in order to use the correct python executable when pip installing and running mypy, I borrowed the relevant env vars from `regenerate_schemas.sh` and ran:
```
$NVIDIA_PYTHON_EXECUTABLE -m pip mypy
$NVIDIA_PYTHON_EXECUTABLE -c 'from mypy import stubgen; stubgen.main()' -m <relevantModule> -v
```
from one dir above the `__init__.py` for `<relevantModule>`. The import/main hack is because I couldn't find any standalone `stubgen` executable as the docs describe.
- `monkeytype` from Instagram - works by running a script and inferring argument types from usage. I tried this quickly with the same `$NVIDIA_PYTHON_EXECUTABLE` on `extension.py` and it failed on these imports:
```
      File "extension.py", line 1, in <module>
    from .bindings import acquire_cesium_omniverse_interface, release_cesium_omniverse_interface, Viewport
ImportError: attempted relative import with no known parent package
```

Again, one or more of these might still work, but this is about as much progress on this issue that I can make for now.

[...]

PR #359, opened 2023-06-20
update cesium omniverse python bindings with new autogenerated stub

mattelser:
partial fix for #309. I've managed to generate this stub mostly automatically, and plan to fully automate the process, but in the mean time it would be useful to have the updated stub available.


mattelser:
closing in favor of a squashed version with `CesiumUsdSchemas` stub as well

PR #360, opened 2023-06-20
update cesium python bindings with new autogenerated stubs

mattelser:
partial fix for #309. Through a partially manual process I can generate python stubs for `CesiumOmniversePythonBindings` and `CesiumUsdSchemas`. I plan to automate the process more fully (and write up the process in the issue linked here) but in the mean-time these improved stubs seem like they would be useful.


mattelser:
fixes #309 completely now. A bash script to generate these stubs automatically came together quicker than expected so I've added it the this PR. It only works in kit 105, but I'm adding it here to prevent divergence. @corybarr can you confirm that your recent changes are represented in these stubs? Thanks!

corybarr:
> fixes #309 completely now. A bash script to generate these stubs automatically came together quicker than expected so I've added it the this PR. It only works in kit 105, but I'm adding it here to prevent divergence. @corybarr can you confirm that your recent changes are represented in these stubs? Thanks!

Odd. I'm still seeing the old schema at line 74 of exts/cesium.usd.plugins/cesium/usd/plugins/CesiumUsdSchemas/__init__.pyi

weegeekps (approved):

genStubs.sh
```diff
@@ -0,0 +1,50 @@
...
+# Get the nvidia python executable path, it has a few modules needed by the source
+# code we're stubbing, so those imports need to be valid
+
+PROJECT_ROOT=`dirname -- "$( readlink -f -- "$0"; )"`
+NVIDIA_USD_BINS="$PROJECT_ROOT/extern/nvidia/_build/target-deps/usd/release/bin"
+NVIDIA_PYTHON_BINS="$PROJECT_ROOT/extern/nvidia/_build/target-deps/python/bin"
+NVIDIA_PYTHON_EXECUTABLE="$NVIDIA_PYTHON_BINS/python3"
+NVIDIA_PYTHON_LIBS="$PROJECT_ROOT/extern/nvidia/_build/target-deps/python/lib"
+NVIDIA_USD_PYTHON_LIBS="$PROJECT_ROOT/extern/nvidia/_build/target-deps/usd/release/lib/python"
+NVIDIA_USDRT_LIBS="$PROJECT_ROOT/extern/nvidia/_build/target-deps/usdrt/_build/linux-x86_64/release/usdrt_only"
```
lilleyse, line 16: `NVIDIA_USDRT_LIBS` can probably be removed now.

genStubs.sh
lilleyse, line 1: Could `regenerate_schema.sh` invoke this script?

lilleyse, line 1: We should also have a batch script for those of us on Windows

genStubs.sh
```diff
@@ -0,0 +1,50 @@
...
+echo "Ensuring mypy is installed"
+$NVIDIA_PYTHON_EXECUTABLE -m pip install mypy
+
+echo "Building lib files flat in temp dir"
+cmake -B build
+cmake --build build
+cmake --install build --component library --prefix $FLAT_LIBRARIES_DIR
+
+# To find the imports mypy has to be run from the same dir as the object files
+cd "$FLAT_LIBRARIES_DIR/lib"
```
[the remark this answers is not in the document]
mattelser: Thanks! Sean pointed out the way to build everything flat, and fwiw it still doesn't work with the USD schema in kit 104

genStubs.sh
mattelser: I don't have access to windows today, should the PR wait on this or can that be a separate issue/PR?

mattelser: Running this does require a fresh build so it could add an annoying amount of time if anyone is iterating on the schema. If that sort of iteration isn't common though I'm happy to add it to `regenerate_schema.sh`

lilleyse: It can be a separate PR

lilleyse: Oh that's true. Maybe the better thing to do is to write a "How to update schemas" guide that we include in the docs folder.

lilleyse: Because in addition to running `regenerate_schemas` and `genStubs` there are other procedures to follow when adding a new attribute or class that can't be automated easily and would be good to have documented somewhere.

lilleyse: Added in https://github.com/CesiumGS/cesium-omniverse/pull/360/commits/20f17e3f5a8b47328726ea9f63c483e078c96371

genStubs.sh
```diff
@@ -0,0 +1,50 @@
+#/bin/bash
```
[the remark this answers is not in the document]
lilleyse: Opened an issue to track this: https://github.com/CesiumGS/cesium-omniverse/issues/366

genStubs.sh
mattelser: opened #367

Issue #355, opened 2023-06-20
More automation for `regenerate_schema`


lilleyse:
After running `./regenerate_schema.(bat|sh)` we need to do some manual fix-up steps

1. Add `CESIUM_API` to all the class declarations
2. Remove `CESIUM_API` from the function declarations
3. Revert "Name" change in `plugInfo.json.in`

It doesn't seem like a lot, but it can be time consuming and it's easy to make mistakes. We should try to automate this.

First we should see if we can remove steps 1 and 2 completely. Kit 105 is using USD 22.08 and aspects of the schema gen or build system may have changed and these manual fix-ups may not be needed. I tried it a few weeks ago and it worked in Windows but not in Linux. Someone else may want to try it in case I messed up somewhere.

For step 3, I'm not sure what the consequences are for changing the plugin name. But we could start by changing the `libraryName` in `cesium_schemas.usda` to `cesiumUsdSchemas` and see what happens.

If any of these fix-up steps are still needed, we should write a Python script to do it for us.

lilleyse:
https://github.com/CesiumGS/cesium-omniverse/pull/363 partially fixed this issue, steps 1 and 2 are no longer required

PR #363, opened 2023-06-22
Regenerate schemas without CESIUM_API manual steps

lilleyse:
Partially addresses https://github.com/CesiumGS/cesium-omniverse/issues/355

Fixes the `regenerate_schemas` step so that we no loner need to manually alter the generated `.h` files. This works for both Kit 104 and 105.

Even though USD doesn't export class definitions for generated schemas the symbol for the class definition still needs to exist in the `.dll`/`.so` otherwise Omniverse will fail to load the extension. The fix is to add `-fvisibility=default` just for the `CesiumUsdSchemas` target. Without the flag, you get runtime errors like:

> 2023-06-21 23:43:09 [60,588ms] [Error] [carb] [Plugin: libcesium.omniverse.plugin.so] Could not load the dynamic library from /home/slilley/Code/cesium-omniverse/exts/cesium.omniverse/bin/libcesium.omniverse.plugin.so. Error: /home/slilley/Code/cesium-omniverse/exts/cesium.omniverse/bin/libcesium.omniverse.plugin.so: undefined symbol: _ZTVN32pxrInternal_v0_22__pxrReserved__13CesiumSessionE (Additional information may be available by running the process with the LD_DEBUG environment variable set)

I didn't mean to tackle this issue, but the solution just clicked during a conversation with @mattelser about https://github.com/CesiumGS/cesium-omniverse/pull/360.

This should fix steps 1 and 2 in https://github.com/CesiumGS/cesium-omniverse/issues/355 but I haven't look into step 3.


corybarr (approved):
Nice. That's a lot of tediousness removed by adding one parameter.

weegeekps (approved):
absolutely-sensational-simon-cowell

PR #372, opened 2023-06-27
Fix c_str bug

lilleyse:
This code is unsafe because key is pointing to memory of a temporary string (use after free bug)

```c++
auto key = std::string(PERSISTENT_SETTINGS_PREFIX).append(USER_ACCESS_TOKEN_PATH).c_str();
return {settings->getStringBuffer(key)};
```

Changed to

```c++
auto key = std::string(PERSISTENT_SETTINGS_PREFIX).append(USER_ACCESS_TOKEN_PATH);
return {settings->getStringBuffer(key.c_str())};
```


weegeekps (approved):
:rocket:

PR #369, opened 2023-06-26
Standardize CesiumUsdSchemas library naming

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/355

PR #374, opened 2023-06-27
Capitalize CesiumUsdSchemas in more places

lilleyse:
Missed a few `cesiumUsdSchemas` -> `CesiumUsdSchemas` changes in https://github.com/CesiumGS/cesium-omniverse/pull/369, in particular in `genStubs`.

https://github.com/CesiumGS/cesium-omniverse/pull/360 and https://github.com/CesiumGS/cesium-omniverse/pull/369 got merged around the same time which is probably why some of these got overlooked.


mattelser (approved):
lgtm

PR #376, opened 2023-06-28
Use clangd for IntelliSense in VSCode on Windows

lilleyse:
`ms-vscode.cpptools` IntelliSense is too slow, switching to clangd instead. We've been using clangd on Linux for a while now without any trouble.


mattelser (approved):
🚀

lilleyse:
I accidentally merged https://github.com/CesiumGS/cesium-omniverse/pull/377 first which is why those commits are showing up in this branch. For future reference, only https://github.com/CesiumGS/cesium-omniverse/pull/376/commits/264cb641b87798f1d74931453e8f1e3ed5d7c0a9 is supposed to be part of this PR.

PR #377, opened 2023-06-28
clang-tidy fixes

lilleyse:
Follow up to https://github.com/CesiumGS/cesium-omniverse/pull/376 - the clangd extension shows clang-tidy errors so I went ahead and fixed them.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,4 +1,4 @@
 ---
-Checks: "-*,bugprone-*,performance-*,modernize-*,-modernize-use-trailing-return-type"
+Checks: "-*,bugprone-*,performance-*,modernize-*,-modernize-use-trailing-return-type,-bugprone-easily-swappable-parameters,-bugprone-exception-escape"
 WarningsAsErrors: "*"
 FormatStyle: none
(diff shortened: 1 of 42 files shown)


.clang-tidy
mattelser, line 1: thank you for removing "easily swappable parameters"!

src/core/include/cesium/omniverse/OmniTileset.h
```diff
@@ -11,14 +11,12 @@
 #include <vector>
 
 namespace Cesium3DTilesSelection {
-struct ImageCesium;
```
mattelser, line 14: Is this removed because `ImageCesium` is available elsewhere?

mattelser (approved):
assuming that the `ImageCesium` comment above is harmless, lgtm

src/core/include/cesium/omniverse/OmniTileset.h
lilleyse: Yeah, other files are including / forward declaring `ImageCesium`. At one point `OmniTileset.h` must have been using it.

Issue #364, opened 2023-06-23
Clipping bug with gltf/pbr.mdl


lilleyse:
After switching from `OmniPBR.mdl` to `gltf/pbr.mdl` in https://github.com/CesiumGS/cesium-omniverse/pull/320 we noticed that distant tiles get clipped early even if the camera's clipping range is set to a very large value. This only happens in the RTX Real-Time renderer and does not happen in Kit 105.

Notice how the distant mountains are clipped in the left screenshot:

RTX Real-Time (Kit 104.2) | RTX Interactive (Kit 104.2)
--|--
![clipping-real-time-104](https://github.com/CesiumGS/cesium-omniverse/assets/915398/e0602dad-3adc-4bcf-a615-c2065354dd3b)|![clipping-interactive-104](https://github.com/CesiumGS/cesium-omniverse/assets/915398/32ecd65a-88e1-4360-bdf5-fdd6083e2585)

It gets worse when you try to zoom out and see the whole earth.

![clipping](https://github.com/CesiumGS/cesium-omniverse/assets/915398/76ac3df2-8888-47c3-a01e-8fa27f34032d)

USD file for testing: [clipping.zip](https://github.com/CesiumGS/cesium-omniverse/files/11849366/clipping.zip)

Issue #335, opened 2023-06-09
stray white tile that won't deload


mattelser:
possibly related to #334. @corybarr first noticed this yesterday and I was able to reproduce it today on linux and windows.
![Screenshot from 2023-06-09 17-12-41](https://github.com/CesiumGS/cesium-omniverse/assets/62680156/308c8b91-e894-4f9e-9f8c-aa47cb5cd4b0)

Cory suggested 
> Dangling tile references maybe? I wonder if they're in Fabric because nothing in the USD stage disables it.

Notably, as can be seen in the image above, the tile does not deload when zoomed in on.

lilleyse:
I can't reproduce this in Kit 105 during normal usage. I think because we're use a geometry pool there and never actually destroying prims.

However I can trigger this 100% of the time when clicking "Disable Geometry Pool" in the `/Cesium` settings. This clears the pools and destroys the fabric prims, leaving white tiles behind. Interestingly, the white tiles are those that were previously visible.

I don't think we have a leak because if you comment out the "Repopulate the asset registry" block in `reloadStage` and click "Print Fabric stage" in the Cesium Debugging window, the stage doesn't doesn't have any pool-related prims.


https://github.com/CesiumGS/cesium-omniverse/assets/915398/ab80ffa3-c1b4-4aa7-aea5-3f23a2ebf826

lilleyse:
I'm pretty sure the issue above was caused by me removing the `/TempChangeTracking` code thinking it was no longer needed in Kit 105. That's fixed in https://github.com/CesiumGS/cesium-omniverse/pull/389.

Issue #214, opened 2023-03-13
Some prims in the tileset can become orphaned when tileset is moved


r-veenstra:
I have observed a few scenarios where moving a dataset can cause prims to become "orphaned" from their tileset.

Eg. they remain in the scene visually, even if the tileset is set to invisible or deleted, or they remain in place when the tileset is moved.

See below, all of those pieces of terrain are now "detached" from the main terrain (which has been deleted from the scene)

![image](https://user-images.githubusercontent.com/123468416/224582753-b96ae4ae-d53a-455b-a658-84e24af21a51.png)

lilleyse:
As a workaround we disabled Fabric geometry streaming, but we should consider re-enabling it when this issue is fixed in a future version of Kit.

lilleyse:
This is working in the most recent Kit pre-release and should ship in the next official release. At that point we can set `geometryStreaming` back to true, if we want.

PR #382, opened 2023-06-30
Upgrade to Kit 105.0.1 and drop support for Kit 104.2

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/364
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/335
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/255
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/214
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/179

Now that USD Composer 2023.1.0 (Beta) is out we can switch to Kit 105. Some of the major changes include:

* Upgrade to USD 22.11
* Upgrade to Python 3.10
* Material ids, values, and primvars can be set dynamically which lets us enable geometry pooling, and later refactor the raster overlay code: [#84](https://github.com/CesiumGS/cesium-omniverse/issues/84) [#242](https://github.com/CesiumGS/cesium-omniverse/issues/242)
* `DynamicTextureProvider` warnings are gone
* Dynamic skies + subsequent adjustments to time of day should now work

Thanks to @jshrake-nvidia, @sirpalee, and @aloysbaillet for their work on this PR.

Let's squash merge this once it's ready to merge since the commit history is pretty long.


lilleyse:
Still need to investigate why the console is being spammed with these messages:
```
2023-06-30 14:29:32 [95,552ms] [Warning] [omni.usd] Coding Error (secondary thread): in _FailGet at line 567 of W:\ac88d7d902b57417\USD\pxr\base\vt\value.cpp -- Attempted to get value of type 'VtArray<GfVec3f>' from VtValue holding 'unsigned __int64'
```

exts/cesium.usd.plugins/schemas/cesium_schemas.usda
```diff
@@ -58,7 +58,7 @@ class CesiumDataPrim "CesiumDataPrim" (
         doc = "Debug option that renders tilesets with textures disabled."
     )
 
-    bool cesium:debug:disableGeometryPool = true (
+    bool cesium:debug:disableGeometryPool = false (
```
corybarr, line 61: This isn't relate to the upgrade. Is there another PR to merge first?

lilleyse: Geometry pooling doesn't work in Kit 104.2 because we can't change a prim's texcoords dynamically. We've wanted to turn it on for a while, and now we can with Kit 105.

corybarr (approved):

.github/workflows/main.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           - {
               name: 'Ubuntu - GCC',
               artifact: 'linux-gcc.tar.xz',
-              os: ubuntu-22.04,
+              os: ubuntu-latest-m,
```
weegeekps, line 31: Hmm, is this something we really want to do? 22.04 is the current LTS, and the one we are all using at Cesium.

weegeekps (commented):
Are there still issues with `usdGenSchema` in Linux?

.github/workflows/main.yml
lilleyse: `ubuntu-latest-m` is still 22.04 I believe.

I needed to switch to a large runner so that there was enough disk space for the kit debug libraries. It became too complicated to download only the release build due to a difference in how USD builds are packaged in 105.

In the near future I think we should switch back to self hosted runners with the nvidia libraries preinstalled.

lilleyse:
> Are there still issues with `usdGenSchema` in Linux?

I don't think so... `regenerate_schema.sh` works fine for me. I'm pretty sure we took care of all the issues in previous PRs.

weegeekps (approved):
This looks good to go.

Issue #242, opened 2023-03-22
Dynamic material support in Fabric


lilleyse:
As of Kit 104.2 it's not possible to update material attributes (including textures) after the material is created or assign new materials to existing prims. So we are stuck with the originally assigned material and any sort of dynamic updates are not possible.

Currently our workaround is to destroy the existing prim whenever material changes happen, e.g. when higher res imagery is attached to the tile.

https://github.com/CesiumGS/cesium-omniverse/blob/0b2499e7faff26dbb357b0260af901e84c721254/src/core/src/FabricPrepareRenderResources.cpp#L158-L163

Once dynamic material updates are supported in Kit, it unlocks several improvements for us:

* No longer need to destroy prims when higher res imagery is attached to the tile
* Mesh and material pooling becomes possible
* Makes it easier to implement multiple raster overlays: https://github.com/CesiumGS/cesium-omniverse/issues/84

CC https://github.com/CesiumGS/cesium-omniverse/pull/169

PR #384, opened 2023-07-01
Material and imagery refactor for Kit 105

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/242

This PR removes a bunch of workarounds that we had in place for Kit 104.2 since it didn't support dynamic material values.

* No longer need to bake material properties like `baseColorFactor`, `emissiveFactor`, etc in the material definition.
* No longer need to bake texcoord translate / scale into the texcoord attribute
* No longer need to destroy prims when higher res imagery is attached to a tile

Only one imagery layer is supported per tileset still. Multiple raster overlays is a bigger lift since it means parsing multiple texcoord sets from the glTF (e.g. geographic vs. web mercator texcoords), storing multiple textures in `FabricMaterial` (possibly requiring a texture pool), and forking gltf/pbr.mdl to read from additional textures.

### DIFF
--- a/src/core/include/cesium/omniverse/FabricGeometry.h
+++ b/src/core/include/cesium/omniverse/FabricGeometry.h
@@ -33,8 +33,6 @@ class FabricGeometry {
         const CesiumGltf::MeshPrimitive& primitive,
         bool smoothNormals,
         bool hasImagery,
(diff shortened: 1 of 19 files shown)


corybarr (approved):
Looks good

PR #388, opened 2023-07-07
Fix USD to GLM matrix conversion

lilleyse:
The USD to GLM matrix conversion functions were assuming USD does right-to-left matrix multiplication, but this is incorrect. On top of that there was a shady `transpose` which cancelled out the mistake in `computeUsdWorldTransform`.

So anything calling `computeUsdWorldTransform` worked fine, but anything calling `glmToUsdMatrix` or `usdToGlmMatrix` directly was broken.

As far as I can tell, this only affects `computeViewState` (fixed in this PR) and`CesiumSession`'s `ecefToUsdTransform` which was incorrectly column-major before, but is now row-major.


weegeekps (approved):
Looks great, works great. :shipit:

Issue #308, opened 2023-05-31
Create unit tests for GltfUtil.cpp


weegeekps:
We need unit tests around `GltfUtil.cpp` as this is a critical piece of the backend code. Doing unit test here will also serve as a sample to the rest of the team and other contributors of how we want our native code unit tested. This may require some refactoring.

PR #389, opened 2023-07-07
More imagery refactoring

lilleyse:
Continuation of the work started in https://github.com/CesiumGS/cesium-omniverse/pull/384.

Added a texture pool that holds onto `DynamicTextureProvider` objects. Textures are now a distinct resource type that can be assigned to multiple materials. This allows multiple tiles to reference the same imagery, which happens when child tiles use their parent's imagery while waiting for higher res imagery. Previously we were copying the imagery due to limitations in Kit 104.

On the performance side, imagery preparation has been offloaded to a worker thread. I think we can do the same for glTF base color textures, I just haven't done it in this PR.

Aside from new files, the main changes are in `FabricPrepareRenderResources` and `FabricMaterial`. `GltfUtil` was refactored heavily to consolidate a bunch of the getters. I'll wait for https://github.com/CesiumGS/cesium-omniverse/pull/387 to be merged first.

As a bonus, this PR adds support for `KHR_texture_transform` because it just neatly fit into the refactoring.

### What's next 

This PR is a stepping stone towards a bigger performance fix. One of the biggest bottlenecks is material compilation, and I think instead of creating a material for each cached tile we can create a material for each selected tile. This will cut the number of materials down from 2000-50000 (depending on your cache size) to around 200-500 (depending on the tileset and your SSE). Any time we allocate lots of materials we get a 10+ second main thread stall, which in my opinion is the most glaring usability issue with the extension right now.

With Kit 105 and the refactoring in this PR we should be able to make a material pool for selected tiles only and update texture paths dynamically.


lilleyse:
I added a few more commits to this PR:

* https://github.com/CesiumGS/cesium-omniverse/pull/389/commits/52b0760938eaa53f5f9cc280551f8cf1820d6191: some cleanup that will be helpful for the next PR
* https://github.com/CesiumGS/cesium-omniverse/pull/389/commits/34e6e3ff6917ef59d2786540fe4a5ac6f5e9f7f6: I realized that we no longer use the `tileId` attribute in Fabric so I removed it
* https://github.com/CesiumGS/cesium-omniverse/pull/389/commits/d2c2a9b924b2049ca2e11e2dbaf39836a876e3bd: moved base color texture loading to the worker thread

lilleyse:
This is ready for review now. I fixed the merge conflict with #308.

src/core/src/FabricGeometry.cpp
```diff
@@ -203,24 +195,27 @@ void FabricGeometry::reset() {
...
-    *tileIdFabric = -1;
+    *worldVisibilityFabric = DEFAULT_VISIBILITY;
     *localToEcefTransformFabric = DEFAULT_MATRIX;
     *worldPositionFabric = DEFAULT_POSITION;
     *worldOrientationFabric = DEFAULT_ORIENTATION;
     *worldScaleFabric = DEFAULT_SCALE;
+    displayColorFabric[0] = DEFAULT_VERTEX_COLOR;
+    displayOpacityFabric[0] = DEFAULT_VERTEX_OPACITY;
 
+    FabricUtil::setTilesetId(_pathFabric, -1);
```
weegeekps, line 216: NIT: Might be worthwhile to explain in a comment or better yet, function documentation for `setTilesetId` as to what `-1` signifies.

weegeekps (approved):
Approved. There's one documentation related nitpick. If you fix it, don't worry about re-approval. I still approve.

PR #391, opened 2023-07-11
Initial Cesium Material Graph nodes

r-veenstra:
This PR contains initial material graph nodes that will eventually allow users to create custom materials for tilesets using the built in Material Graph in USD Composer.  It also includes a "Cesium GLTF Material" that directly inherits from the regular GLTF material. 
 Currently some manual configuration is required to make these work until we seek feedback on how to automate this.

To activate, go to Preferences > Material and add the mdl folder from the cesium.omniverse extension into the Custom Paths list

image

Then go to User Allow List and add cesium.mdl to the list

image

Restart USD Composer and you should now have access to these nodes in the Material Graph

image

Right click in the stage list to create a new material and choose Add MDL File.  Find mdl/cesium.mdl in the cesium.omniverse extension.  Inside this is a material that directly inherits from GLTF and at the moment, does not include any other changes (but could at a later date)

image

Drag the new material onto the default ground prim on the stage.  Note the material will be duplicated under the "Looks" scope of the stage.  Right click the new material and choose Open in MDL Material Graph.  It will look like this

image

Drag a "World-mapped texture lookup float3" node onto the graph.  Choose an image, then connect the nodes output into a "color from float 3" node, then connect that to the base color factor.  It should look like the below:

image

You should see the texture mapped to the ground, note it will be using world space UV's.  You can adjust the UV mapping in the lookup_world_texture_float3 node.  No matter where you move the ground prim, its UV mapping should remain in world space.

image

And that's it.  I'm currently experiencing an issue with my commented out lookup_world_texture_color, which is just a variant of the same function but returns color instead of float3.  There is some sort of incompatibility with the code used in the world UV mapping that needs to be further understood before it can be used.


lilleyse:
@r-veenstra I pushed https://github.com/CesiumGS/cesium-omniverse/pull/391/commits/9f3f5a128b51d414481f14964dcc43f7d971dc65 so that we don't have to do all the setup stuff when testing things locally. We may be able to set some of those settings programmatically so that it works out of the box in USD Composer.

apps/cesium.omniverse.dev.kit
```diff
@@ -17,6 +18,15 @@ app.useFabricSceneDelegate = true
...
 app.fastShutdown = true
+# Both searchPaths settings must be set in order for material graph to find cesium mdl exports
+renderer.mdl.searchPaths.custom = "${app}/../exts/cesium.omniverse/mdl"
+materialConfig.searchPaths.custom = ["${app}/../exts/cesium.omniverse/mdl"]
+materialConfig.materialGraph.userAllowList = ["cesium.mdl"]
 
 [settings.app.exts]
-folders.'++' = ["${app}/../exts"] # Make extensions from this repo available.
+folders.'++' = [
+    "${app}", # Find other applications in this folder
```
weegeekps, line 28: Why is this change required? I'm not sure I follow.

weegeekps (commented):
Code looks good, but is this supposed to work with the Cesium Dev App? We will need tests for it eventually, so it does need to work within that.

apps/cesium.omniverse.dev.kit
r-veenstra: @weegeekps Sean implemented this as part of changes to make the Material Graph Nodes available inside of the Cesium Dev App (which I can confirm is working)

Based on the other changes in that commit, it looks like he has moved a handful of duplicate settings that were repeated across our other apps into the main cesium.omniverse.dev.kit, which makes sense as our other apps use cesium.ominverse.dev.kit as a dependency.

weegeekps:
@r-veenstra I'm not sure what I'm doing wrong, but I don't see the Cesium functions in the Material Graph editor for this PR.

r-veenstra:
> @r-veenstra I'm not sure what I'm doing wrong, but I don't see the Cesium functions in the Material Graph editor for this PR.

OK I've managed to repro - will dig a bit deeper, but essentially the search path for the MDL is being set correctly based on Sean's change, but the app isn't picking it up unless it has been manually saved into material.config.toml.  It would have been working on my end initially as I already had this path saved.

lilleyse:
I believe it's a Fabric population bug. I noticed the problem https://github.com/CesiumGS/cesium-omniverse/pull/394 as well:

> There seems to be a Fabric population bug where assigning a material doesn't work unless you enable fabric scene delegate after load. So in order to test this with `cesium.omniverse.dev` you'll need to set `app.useFabricSceneDelegate = false`.

r-veenstra:
> I believe it's a Fabric population bug. I noticed the problem #394 as well:
> 
> > There seems to be a Fabric population bug where assigning a material doesn't work unless you enable fabric scene delegate after load. So in order to test this with `cesium.omniverse.dev` you'll need to set `app.useFabricSceneDelegate = false`.

I just tried setting FSD to false, however it still would error out in the same way when the Material Graph is trying to load the Cesium functions.  Only once I saved the path into material.config.toml, did it work.

lilleyse:
Weird, it seems to work for me and I confirmed that I don't have a `material.config.toml` file in `Documents\Kit\shared`

lilleyse:
@r-veenstra is it possible to make this work with both y-up and z-up stages? Like maybe `max_world_xy` / `min_world_xy` should actually be `max_world_xyz` / `min_world_xyz`?

With a y-up stage it basically textures the wrong axis.

image

[simple-test.zip](https://github.com/CesiumGS/cesium-omniverse/files/12221930/simple-test.zip)

exts/cesium.omniverse/mdl/cesium.mdl
```diff
@@ -0,0 +1,77 @@
...
+import ::tex::*;
+
+using ::gltf::pbr import *;
+
+module [[
+    anno::version( 1, 0, 0),
+    anno::display_name("Cesium MDL functions")
+]];
+
+export float2 world_coordinate_2d(float2 min_world_xy=float2(-5000.0,-5000.0), float2 max_world_xy=float2(5000.0,5000.0))
```
lilleyse, line 14: The default values don't seem to have any effect. They default to 0 and I have to go in and edit them separately.

lilleyse:
`lookup_world_texture_color` seems to be working fine for me after adding `wrap_u` and `wrap_v`. I added it in https://github.com/CesiumGS/cesium-omniverse/pull/391/commits/10a5bc9b77448cc4aa37864ac63ae59ac8e6409a

image

[cesium-nodes.zip](https://github.com/CesiumGS/cesium-omniverse/files/12222338/cesium-nodes.zip)

exts/cesium.omniverse/mdl/cesium.mdl
lilleyse: Oh weird, I can't reproduce this anymore. Maybe I had Fabric enabled by default and it didn't populate into USD correctly?

lilleyse:
I'm going to close this PR so that branch management is easier. All these commits are included in #406

PR #393, opened 2023-07-12
Update styling of credits

r-veenstra:
I noticed the styling of credits with logos could be improved.  Below is the previous state - inconsistent image heights, items not aligned to the bottom of window, etc.

credits

I've made two main modifications.  Firstly, each credit, along with the data attribution button, have been wrapped in individual VStacks.  A VSpacer has been added as the first item in the stack, followed by the button/image/text.  This will ensure the credit item is always pushed to the bottom of the VStack, regardless of the height of the parent container.  The results is credits can be differing heights but will always be pushed to the bottom of the screen, the preferred outcome.

You'll see below each credit (green box) has a VSpacer above, ensuring the VStack (orange boxes) are filled to the height of the largest credit.

image

Secondly, I've updated CesiumImageButton to support an optional height parameter.  This allows us to enforce a height for image credits, regardless of the size of the source image.  This does have some potential side effects if users add credits to their datasets in ion that are particularly large or oddly sized, however this is probably a wider issue across all runtimes.  We don't necessarily need to implement this now and it would be good to get some feedback, but I do think it makes sense to restrict or unify the sizing

I've also fixed some minor padding/margin issues with CesiumImageButton where the image wasn't being properly centered in the button.

Lastly I've updated CesiumUriImage to use similar logic, also removing an odd recursive bug that was referenced in the previous code.  

The updated style is shown below below

credits_final


lilleyse:
I've leave the review to @weegeekps, but the approach makes sense to me and the screenshots look great.

weegeekps (approved):
Looks great. :shipit:

PR #395, opened 2023-07-12
Print connections for "Print Fabric stage" button

lilleyse:
This lets us see `connection` values when we print the fabric stage, e.g.

```
Prim: /World/Looks/cesium_gltf_material/Shader (12583937)
  Attributes:
    Attribute: inputs:base_color_texture
      Type: connection
      Value: Path: /World/Looks/cesium_gltf_material/cesium_base_color_texture, Attribute Name: outputs:out
```

This will come in handy for debugging tileset materials.

### DIFF
--- a/src/core/src/FabricUtil.cpp
+++ b/src/core/src/FabricUtil.cpp
@@ -139,6 +139,19 @@ std::string printAttributeValue(
     }
 }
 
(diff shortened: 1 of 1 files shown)


weegeekps (approved):
LGTM

PR #396, opened 2023-07-14
Abstracted out wheel installation

weegeekps:
This PR abstracts out the vendor wheel installation a bit, so we can have each extension handle it's own wheels.

As an example, I also went ahead and added `pyproj` and it's dependency to `cesium.powertools`. Dependencies just need to be included, and don't necessarily need to be manually installed.

This has been tested in both Linux and Windows, but it would be good if other's could test this on their systems as well.


exts/cesium.powertools/vendor/certifi-2023.5.7-py3-none-any.whl
lilleyse, line 1: What is this file for?

lilleyse:
@weegeekps just that one comment from me, looks great

exts/cesium.powertools/vendor/certifi-2023.5.7-py3-none-any.whl
weegeekps: This is a dependency for pyproj. It's just the Mozilla certs we already bundle a different way.

I looked at potential methods for replacing what we are doing with this and came up short. The way this package works will only work for Python.

lilleyse (approved):

PR #397, opened 2023-07-18
Fixed ECEF to USD transform calculation

weegeekps:
We were doing the ECEF to USD maths wrong which was causing us issues when we tried to do the inverse USD to ECEF.

```cpp
const auto primUsdWorldTransform = computeUsdWorldTransform(primPath);
```

This is actually giving us the Local to World transformation, rather than what we needed, the World to Local transformation. Inverting it gives us what we want.

Additionally, I swapped over to using `affineInverse` for the slight speed benefit. Doubt it will matter but we'll see.

### DIFF
--- a/src/core/src/UsdUtil.cpp
+++ b/src/core/src/UsdUtil.cpp
@@ -1,12 +1,14 @@
 #include "cesium/omniverse/UsdUtil.h"
 
 #include "cesium/omniverse/Context.h"
+#include "cesium/omniverse/GeospatialUtil.h"
 #include "cesium/omniverse/Viewport.h"
 
 #include <CesiumGeometry/Transforms.h>
 #include <CesiumGeospatial/Cartographic.h>
 #include <CesiumGeospatial/GlobeTransforms.h>
 #include <glm/gtc/matrix_access.hpp>
+#include <glm/gtc/matrix_inverse.hpp>
 #include <glm/gtx/matrix_decompose.hpp>
 #include <omni/ui/ImageProvider/DynamicTextureProvider.h>
 #include <pxr/usd/sdf/primSpec.h>
@@ -20,34 +22,6 @@
 #include <regex>
 
 namespace cesium::omniverse::UsdUtil {
-
-namespace {
-glm::dmat4 getEastNorthUpToFixedFrame(const CesiumGeospatial::Cartographic& cartographic) {
-    const auto cartesian = CesiumGeospatial::Ellipsoid::WGS84.cartographicToCartesian(cartographic);
-    const auto matrix = CesiumGeospatial::GlobeTransforms::eastNorthUpToFixedFrame(cartesian);
-    return matrix;
-}
-
-glm::dmat4 getAxisConversionTransform() {
-    const auto upAxis = getUsdUpAxis();
-
-    auto axisConversion = glm::dmat4(1.0);
-
-    // USD up axis can be either Y or Z
-    if (upAxis == pxr::UsdGeomTokens->y) {
-        axisConversion = CesiumGeometry::Transforms::Y_UP_TO_Z_UP;
-    }
-
-    return axisConversion;
-}
-
-glm::dmat4 getUnitConversionTransform() {
-    const auto metersPerUnit = getUsdMetersPerUnit();
-    const auto matrix = glm::scale(glm::dmat4(1.0), glm::dvec3(metersPerUnit));
-    return matrix;
-}
-
-} // namespace
 
 pxr::UsdStageRefPtr getUsdStage() {
     return Context::instance().getStage();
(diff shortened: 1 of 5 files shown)


src/core/src/UsdUtil.cpp
```diff
@@ -222,25 +196,18 @@ pxr::SdfAssetPath getDynamicTextureProviderAssetPath(const std::string& name) {
...
+        GeospatialUtil::getCoordinateSystem(origin, getUsdMetersPerUnit()).getEcefToLocalTransformation();
+    const auto primInverseUsdWorldTransform = glm::affineInverse(computeUsdWorldTransform(primPath));
+    const auto primEcefToUsdTransform = primInverseUsdWorldTransform * ecefToUsdTransform;
     return primEcefToUsdTransform;
 }
 
 glm::dmat4
 computeUsdToEcefTransformForPrim(const CesiumGeospatial::Cartographic& origin, const pxr::SdfPath& primPath) {
-    return glm::inverse(computeEcefToUsdTransformForPrim(origin, primPath));
+    return glm::affineInverse(computeEcefToUsdTransformForPrim(origin, primPath));
```
corybarr, line 210: Is this an optimization? I see `glm::inverse` in a few other places and wonder if we should use `affineInverse` exclusively. I don't see why we'd ever have any non-affine transformations.

corybarr (approved):
All looks good

src/core/src/UsdUtil.cpp
weegeekps: Yeah, I agree. As far as I'm able to tell, `cesium-native` prefers `affineInverse` as well. Seems like a safe optimization to me.

weegeekps: Also, just to add... This exploits that the matrix is affine to be much faster in its calculation, although I admit I don't have any numbers handy.

lilleyse:
I can see why this order of transformations makes sense in the context of globe anchors, but I don't think changing `computeEcefToUsdTransformForPrim` is the right thing to do.

`computeEcefToUsdTransformForPrim` is meant to go from `ECEF` -> `USD local` -> `USD world` so that stage transforms on the tileset prim are handled correctly. By swapping the order, it causes the transforms to be inverted like shown in this video:

https://github.com/CesiumGS/cesium-omniverse/assets/915398/b9edb25a-89fc-403f-be20-26987cc32a6b

src/core/src/UsdUtil.cpp
lilleyse: `inverse` vs. `affineInverse` could matter if we end up supporting non uniform scale, like stretching the tileset in the up-axis.

I haven't tried Unity or Unreal to see if it's expected to work in native.

weegeekps:
@lilleyse Yup, nice catch. I missed this. Let me play around with a few potential fixes. I think it's plausible that we may be trying to do too much with this one function, and I'm not terribly confident that our nomenclature here makes a lot of sense.

src/core/src/UsdUtil.cpp
lilleyse: Never mind, an affine matrix can have non-uniform scale.

kring:
> computeEcefToUsdTransformForPrim is meant to go from ECEF -> USD local -> USD world so that stage transforms on the tileset prim are handled correctly.

Very good point, we have the same situation in Unity, and are now implementing it in Unreal as well. For exactly the reason you've stated.

There are subtleties, though! And maybe that's where you're running into problems. The transformation is always, as you've said, `ECEF -> USD local -> USD world`. However, the `USD local -> USD World` for a _globe anchored_ object is actually the _parent's_ local->world. It does not and must not include the transformation of the anchored object itself.

Let's say you have an attachment hierarchy like this (sorry, I don't know the Omniverse terminology):

CesiumGeoreference
↳ Cesium World Terrain

kring:
Let's break this down a bit more, as much to make sure I'm thinking about it correctly as anything. Let's say we have this hierarchy:

```
CesiumGeoreference
 ↳ Parent
    ↳ Child (has a CesiumGlobeAnchor)
```

Now in the steady state where nothing is changing, each of these objects has a Transform and they're composed in the normal way. If the user changes the CesiumGeoreference transform, or the Parent transform, or even the Child transform, this will impact their position/orientation/scale in the game engine world exactly in the normal way that users would expect.

But there are two kinds of important changes that might happen:

1. The game engine transform changes: this requires recomputing the ECEF transform from the game engine Transform.
2. The CesiumGeoreference origin changes: this requires recomputing the game engine Transform from the ECEF transform.

## The game engine Transform changes

A CesiumGlobeAnchor maintains an anchored object -> ECEF matrix. In the case of the Child transform to Parent changing, this is computed from the game engine transform very simply:

```cpp
const LocalHorizontalCoordinateSystem& georeference = /*Local Horizontal coordinate system defined by the CesiumGeoreference origin*/
GlobeAnchor anchor = GlobeAnchor::fromAnchorToLocalTransform(georeference, childToParentTransformation)
glm::dmat4 anchorToEcef = anchor.getAnchorToFixedTransform();
```

Note that this does not in any way involve the transformations of Parent to CesiumGeoreference, or of CesiumGeoreference to game world! 

This may seem surprising, but it's intentional. With this setup, by moving the CesiumGeoreference, we move the whole globe in the game world. By moving Parent within CesiumGeoreference, we move a subset of the globe within the game world. It's usually not very useful for an object like Parent to have a non-identity Transform, but it sometimes can be to, for example, offset one photogrammetry dataset to line up with another. Think about the case where Parent is a Cesium3DTileset, and Child is an individual Tile within it.

Of course, I'm still not 100% sure we have this right. This issue explains one way it's not great:
https://github.com/CesiumGS/cesium-unity/issues/132

Offsetting a tileset in terms of game world coordinates is kind of wrong. Lifting Melbourne up in the +Z direction so CWT doesn't poke through it is fine when the origin is in Melbourne. But if the georeference origin changes such that +Z isn't up in Melbourne anymore, that +Z transformation on the Melbourne tileset will become useless or harmful.

But that issue essentially proposes solving this by making it so that a GlobeAnchor on Child affects the transform of Parent when the Georeference changes, which is likely to be even more confusing for users. Maybe for a Tileset it's ok. Maybe. But not for an arbitrary parent.

## The CesiumGeoreference origin changes

Similar to the above, a changing georeference origin only impacts the game engine Transform of objects that have a globe anchor. So the game engine transform on `CesiumGeoreference` and `Parent` stay the same. It looks like this:

```cpp
GlobeAnchor anchor = GlobeAnchor::fromAnchorToFixedTransform(anchorToEcefComputedPreviously);
const LocalHorizontalCoordinateSystem& georeference = /*Local Horizontal coordinate system defined by the CesiumGeoreference origin*/
glm::dmat4 newChildTransform = anchor.getAnchorToLocalTransform(georeference);
```

PR #398, opened 2023-07-24
Temporary workaround(s) to get python tests running

mattelser:
Currently nvidia is looking into an issue where the `pxr` renderer and `usdrt` are not working together correctly. Running python tests crashes without this at the moment.
Of note, one particular test (checking if the main Cesium window is docked) will also fail iff powertools is enabled in `cesium.omniverse.dev.kit`. I am not sure why yet, but would welcome any input on that here or in slack.

### DIFF
--- a/exts/cesium.omniverse/config/extension.toml
+++ b/exts/cesium.omniverse/config/extension.toml
@@ -54,8 +54,8 @@ persistent.exts."cesium.omniverse".userAccessToken = ""
 
 [[test]]
 args = [
-    "--/renderer/enabled=pxr",
-    "--/renderer/active=pxr",
+    "--/renderer/enabled=rtx",
+    "--/renderer/active=rtx",
     "--/app/window/dpiScaleOverride=1.0",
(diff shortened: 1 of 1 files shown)


weegeekps:
@mattelser I'm wondering if the power tools window is taking attention away from the Cesium window, causing that test to fail?

weegeekps:
Also, do we even want to run the tests with power tools enabled?

mattelser:
> 

This was it. calling `focus()` on the main cesium window makes the dock test (not to be confused with `doctest`) pass. However, now it hangs when starting to exit.

mattelser:
We may very well want to disable powertools for tests at some point, but the call to `focus()` does fix the only test that was the problem. As for the hangs I ran into, they were an old setting ("--no-window") that I had left over from prior debugging. The most recent commit means this PR would allow python tests to work, though they currently only work via the `run_python_unit_tests.(sh|bat)` shell command. I'm opening an issue for running them through the app

mattelser:
closing in favor of #401

PR #402, opened 2023-07-26
Fix Z up

### LINKED ISSUE #400: Tilesets are incorrectly oriented with stage default up axis set to Z
When using a stage default up axis set to Z, tiles load with an incorrect orientation.

Screenshot_10

I've confirmed this is a regression from changes in https://github.com/CesiumGS/cesium-omniverse/pull/397.  Prior to this both stage options worked ok.

Repro steps are to simply set the Stage Default Up Axis to Z, create a new stage, and add the Cesium World Terrain.

weegeekps:
Resolves #400.

I inadvertently removed the code that handles z+ up when I switched us to use the `LocalHorizontalCoordinateSystem` in Cesium Native.

This also reverts the affine inverse I also added in #397 that broke the gizmo for the tileset. There's a better solution for that problem that I will be implementing soon.

### DIFF
--- a/src/core/src/GeospatialUtil.cpp
+++ b/src/core/src/GeospatialUtil.cpp
@@ -1,5 +1,7 @@
 #include "cesium/omniverse/GeospatialUtil.h"
 
+#include "cesium/omniverse/UsdUtil.h"
+
 namespace cesium::omniverse::GeospatialUtil {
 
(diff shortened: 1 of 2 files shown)


weegeekps:
@r-veenstra This PR should fix Z+ up for your AEC stuff. Sorry for breaking this.

r-veenstra (approved):
I've tested both Y up and Z up configurations, including testing against imported spatial data from 3DS max.  Confirmed CWT loads in the correct orientation compared to the exported spatial data.  Looks good with me!

PR #403, opened 2023-07-26
visually enable/disable top bar buttons based on sign-in status

mattelser:
Previously all buttons would appear clickable even if they were deactivated. This greys out the correct buttons in the top bar of the Cesium window when the user is signed out and enables them when the user is signed in.
Screenshot from 2023-07-26 15-57-47
Screenshot from 2023-07-26 15-58-07

### DIFF
--- a/exts/cesium.omniverse/cesium/omniverse/ui/styles.py
+++ b/exts/cesium.omniverse/cesium/omniverse/ui/styles.py
@@ -44,6 +44,15 @@ class CesiumOmniverseUiStyles:
         "Button.Label": {"alignment": Alignment.CENTER_BOTTOM},
     }
 
+    top_bar_button_disabled_style = {
+        "Button": {"padding": 10.0, "stack_direction": Direction.TOP_TO_BOTTOM},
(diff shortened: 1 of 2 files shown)


exts/cesium.omniverse/cesium/omniverse/ui/styles.py
```diff
@@ -44,6 +44,15 @@ class CesiumOmniverseUiStyles:
...
 
+    top_bar_button_disabled_style = {
+        "Button": {"padding": 10.0, "stack_direction": Direction.TOP_TO_BOTTOM},
+        "Button.Image": {
+            "alignment": Alignment.CENTER,
+            "color": cl("#808080"),
+        },
+        "Button.Label": {"alignment": Alignment.CENTER_BOTTOM, "color": cl("#808080")},
+    }
+
```
weegeekps, line 55: Did you check if there is a state selector for this? I believe you should be able to use `Button:disabled` as the selector.

weegeekps (approved):
LGTM

PR #404, opened 2023-07-27
Powertools for syncing Sun Study with CesiumGeoreference

r-veenstra:
Two small additions to powertools here:

1) A simple button in the Powertools menu to set Sun Study latitude/longitude based on the current CesiumGeoreference latitude/longitude.  This also ensures the north angle is set to 90 degrees, which is required to ensure the sun is at the correct yaw compared to the globe.

2) An addition to the Georeference Helper which will extract latitude/longitude from the Environment prim and set the CesiumGeoreference to match.  This is useful as some of the Omniverse connectors such as Rhino + Sketchup store the latitude and longitude values from the export in environment prim.

### DIFF
--- a/exts/cesium.powertools/cesium/powertools/utils/utils.py
+++ b/exts/cesium.powertools/cesium/powertools/utils/utils.py
@@ -1,6 +1,6 @@
 import omni.usd
 from omni.kit.viewport.utility import get_active_viewport
-from pxr import Gf, UsdGeom
+from pxr import Gf, UsdGeom, Sdf
 import json
 import carb.settings
 import os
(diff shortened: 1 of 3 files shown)


weegeekps (approved):
We're starting to get some really useful tools in power tools. Maybe we should start looking into a way to actually package power tools as an optional package?

exts/cesium.powertools/cesium/powertools/utils/utils.py
```diff
@@ -47,3 +47,27 @@ def save_carb_settings(powertools_extension_location: str):
...
+    cesium_prim = stage.GetPrimAtPath("/CesiumGeoreference")
+
+    lat_attr = get_or_create_attribute(environment_prim, "location:latitude", Sdf.ValueTypeNames.Float)
+    lat_attr.Set(cesium_prim.GetAttribute("cesium:georeferenceOrigin:latitude").Get())
+
+    long_attr = get_or_create_attribute(environment_prim, "location:longitude", Sdf.ValueTypeNames.Float)
+    long_attr.Set(cesium_prim.GetAttribute("cesium:georeferenceOrigin:longitude").Get())
+
+    north_attr = get_or_create_attribute(environment_prim, "location:north_orientation", Sdf.ValueTypeNames.Float)
+    north_attr.Set(90.0)  # Always set to 90, otherwise the sun is at the wrong angle
```
lilleyse, line 73: Just checking, is this fix needed for both y-up and z-up stages?

r-veenstra: Yes, I checked both when implementing.  NVIDIA also bakes the 90 degree rotation into all of their exports via their connectors from AEC software.

lilleyse (approved):

PR #405, opened 2023-07-31
Remove rtx_plugins dependency

lilleyse:
We no longer depend on the `rtx_plugins` package, so I removed it from our build system. I also renamed the usdrt target to fabric.

There may be more changes needed once Kit 105.1 is out, but this is an incremental step towards that.


weegeekps:
@lilleyse Is there anything special we should try, such as deleting certain folders, when reviewing this?

weegeekps (approved):
:shipit:

PR #407, opened 2023-08-01
Make show on startup a configurable setting

r-veenstra:
Some integrations of Cesium for Omniverse do not require a user interface to appear. Currently the interface appears by default and this cannot be controlled on an per app basis.

The existing showOnStartup parameter has been connected to an Omniverse setting of the same name, which enables this behavior to be configured on a per application basis.  

Simply add the following to your applications kit file to prevent the UI from displaying by default. 

```
[settings.exts."cesium.omniverse"]
showOnStartup = false
```

I have tested this across USD composer and view.  Default behavior should be to show the UI, so users should not experience anything different unless manually configuring this setting.

### DIFF
--- a/exts/cesium.omniverse/cesium/omniverse/extension.py
+++ b/exts/cesium.omniverse/cesium/omniverse/extension.py
@@ -26,7 +26,6 @@ from .ui.credits_viewport_controller import CreditsViewportController
 
 cesium_extension_location = os.path.join(os.path.dirname(__file__), "../../")
 
-
 class CesiumOmniverseExtension(omni.ext.IExt):
     @staticmethod
     def _set_menu(path, value):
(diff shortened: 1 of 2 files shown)


lilleyse:
Doesn't have to hold up this PR and probably more a question for @weegeekps - should other Cesium windows appear in `Window -> Cesium` or is it intentionally limited to just the debugging window?

Untitled

lilleyse (approved):

r-veenstra:
> Doesn't have to hold up this PR and probably more a question for @weegeekps - should other Cesium windows appear in `Window -> Cesium` or is it intentionally limited to just the debugging window?
> 

Good find, I'd expect all three (Cesium, Assets and Debugging) to be present.  Currently this means if I close the Cesium UI, I can't actually bring it back and need to restart my app.

I'm sure at some point all menu options were available, so this could be a bug unless functionality has intentionally changed for another reason.  I've checked main prior to this PR and it is still behaving the same way, so I don't think this PR has introduced it.  I'll create an issue to cover this.

r-veenstra:
I've added an issue [here ](https://github.com/CesiumGS/cesium-omniverse/issues/408)as I could repro at least as far back as release 0.8.0

PR #409, opened 2023-08-01
Fixed the windows not showing up in the menu

### LINKED ISSUE #408: Cesium for Omniverse menu items are missing
In release 0.8.0 and later, the Cesium UI and Cesium Assets UI menu items are missing from the Window > Cesium menu

The main implication is users cannot recover the Cesium UI if it has been closed by the user

image

weegeekps:
Resolves #408.

I believe the signature of the `add_item` option in `omni.kit.ui.Menu` has changed slightly to this new signature:

`def add_item(self, menu_path: str, on_click: typing.Callable[[str, bool], None] = None, toggle: bool = False, priority: int = 0, value: bool = False, enabled: bool = True, original_svg_color: bool = False, auto_release: bool = True) -> carb._carb.Subscription: ...`

I don't remember seeing the SVG color or the `auto_release` arguments, and in fact it was the `auto_release` in addition to the fact that we weren't properly keeping track of our menus here that was hurting us. Since we kept overwriting `self._menu` with whatever the last menu item was, the other two options were being released and thus not showing.

I fixed this by turning `self._menu` into an array named `_menus` and storing our menu items in it.

### DIFF
--- a/exts/cesium.omniverse/cesium/omniverse/extension.py
+++ b/exts/cesium.omniverse/cesium/omniverse/extension.py
@@ -55,7 +55,7 @@ class CesiumOmniverseExtension(omni.ext.IExt):
         self._attributes_widget_controller: Optional[CesiumAttributesWidgetController] = None
         self._credits_viewport_controller: Optional[CreditsViewportController] = None
         self._logger: logging.Logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


lilleyse (approved):
Works nicely

PR #414, opened 2023-08-02
Release FabricMeshes stored in TileLoadThreadResult

lilleyse:
If a tile made it through `prepareInLoadThread` but not `prepareInMainThread` its fabric meshes would not be released back into the pool, causing a memory leak.

The bug was introduced in https://github.com/CesiumGS/cesium-omniverse/pull/389 - fabric meshes were added to the `TileLoadThreadResult` struct but they were not being released in `free`.

You can see the fix in action by adding a tileset, removing the tileset, and looking at the pool capacity. Previously the pool capacity would not return back to 2048.

Before|After
--|--
before-tileset-destroyed|after-tileset-destroyed

### DIFF
--- a/src/core/src/FabricPrepareRenderResources.cpp
+++ b/src/core/src/FabricPrepareRenderResources.cpp
@@ -193,6 +193,28 @@ void setFabricMeshes(
     }
 }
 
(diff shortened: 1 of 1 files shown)


mattelser (approved):
LGTM

weegeekps (approved):
LGTM

PR #415, opened 2023-08-02
remove thin USD wrappers from C++ API in favor of underlying USD functions

### LINKED ISSUE #390: Remove functions from C++ API
There are several functions in the C++ API that are thin wrappers around adding or removing prims from the USD stage. We should remove these functions from `CesiumOmniverse.h/cpp` and `Context.h/cpp` and use the Python USD API directly.

These functions can be removed to start:

* `addTilesetUrl`
* `addTilesetIon`
* `addImageryIon`
* `addTilesetAndImagery`
* `isTileset`
* `removeTileset`

For reference, look at how `cesium.performance.app` adds and removes tilesets without touching the C++ API.

mattelser:
fixes #390 . I didn't actually see these functions referenced anywhere, so this only required removal and not any replacement. It builds and the app runs, I'm not sure if there's any additional specific test cases I should try


weegeekps:
These used to be used in the debugging menu. I'm unsure if they still have value for debugging needs. I haven't used them in quite a while.

@lilleyse We may want to put this on the agenda for stand-up today.

lilleyse:
Most of these functions are still referenced in the python code as snake_case, e.g. look for `add_tileset_ion`

weegeekps:
@mattelser Can you close this or convert it to a draft so it doesn't get approved or merged while you're fixing things?

PR #416, opened 2023-08-02
Better handling for tilesets that are destroyed mid load

lilleyse:
After talking with @kring and reading through https://github.com/CesiumGS/cesium-native/pull/554 I realized that `FabricPrepareRenderResources` had a few incorrect assumptions about tileset lifetime management.

When a tileset is destroyed its `TileContentManager` still exists to finish processing tiles that were loading asynchronously. So we need to do a few things:

1. Be careful not to access a tileset that's already been destroyed. Previously we were storing a reference to the tileset in `FabricPrepareRenderResources`, but this has the risk of becoming a dangling reference. Now we store a pointer and set it to null when the tileset is destroyed.
2. Ensure that asynchronously loading tiles make it through `prepareInLoadThread` and `prepareInMainThread` so that the fabric resources can be unloaded later. The previous code would return early a bit too aggressively and potentially lead to memory leaks. A couple `tilesetExists()` checks have been removed.
3. Since tiles may be processed after the fabric stage has been destroyed, make sure we don't modify fabric prims that no longer exist. We had some code like this already, but the new code is more robust.

### DIFF
--- a/src/core/src/FabricGeometry.cpp
+++ b/src/core/src/FabricGeometry.cpp
@@ -34,25 +34,43 @@ const auto DEFAULT_VISIBILITY = false;
 FabricGeometry::FabricGeometry(
     const pxr::SdfPath& path,
     const FabricGeometryDefinition& geometryDefinition,
-    bool debugRandomColors)
+    bool debugRandomColors,
+    long stageId)
     : _pathFabric(path.GetText())
     , _geometryDefinition(geometryDefinition)
-    , _debugRandomColors(debugRandomColors) {
+    , _debugRandomColors(debugRandomColors)
+    , _stageId(stageId) {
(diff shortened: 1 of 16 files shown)


src/core/src/FabricGeometry.cpp
```diff
@@ -181,10 +203,6 @@ void FabricGeometry::initialize() {
 }
 
 void FabricGeometry::reset() {
-    if (!UsdUtil::hasStage()) {
-        return;
-    }
-
```
weegeekps, line 187: Shouldn't the `stageDestroyed` guard go here?

lilleyse: It only needs to go in public member functions. `setActive` will do this check before it calls `reset`.

PR #417, opened 2023-08-02
Remove empty pools

lilleyse:
Fixes one of the items in https://github.com/CesiumGS/cesium-omniverse/pull/406:

> Delete material pools that are no longer in use. This is especially important now that the tileset material is part of the material definition.

Basically we don't want to keep empty pools around because the inactive prims just take up space and may never be reused.


weegeekps (approved):
LGTM

PR #420, opened 2023-08-04
Remove cudart from build system

lilleyse:
Similar to https://github.com/CesiumGS/cesium-omniverse/pull/405, we no longer depend on cudart due to internal restructurings in Kit 105.

This fixes a problem where the extension doesn't load in Code 2023.1.1:

> 2023-08-04 14:11:42  [Error] [carb] [Plugin: libcesium.omniverse.plugin.so] Could not load the dynamic library from /cesium-omniverse/exts/cesium.omniverse/bin/libcesium.omniverse.plugin.so. Error: libcudart.so.11.0: cannot open shared object file: No such file or directory (Additional information may be available by running the process with the LD_DEBUG environment variable set)

We'll need to figure this out eventually since @corybarr is using cuda for point clouds, but for now we just need to get things working with Code 2023.1.1.


weegeekps (approved):
Works okay for me.

PR #426, opened 2023-08-15
Material cleanup

lilleyse:
This PR contains a lot of the cleanup related changes from https://github.com/CesiumGS/cesium-omniverse/pull/406.

I wanted to separate into it's own PR so that https://github.com/CesiumGS/cesium-omniverse/pull/406 is easier to review.

### DIFF
--- a/src/core/src/FabricMaterial.cpp
+++ b/src/core/src/FabricMaterial.cpp
@@ -1,41 +1,31 @@
 #include "cesium/omniverse/FabricMaterial.h"
 
+#include "cesium/omniverse/Context.h"
 #include "cesium/omniverse/FabricAttributesBuilder.h"
-#include "cesium/omniverse/FabricMaterialDefinition.h"
-#include "cesium/omniverse/FabricTexture.h"
 #include "cesium/omniverse/FabricUtil.h"
-#include "cesium/omniverse/GltfUtil.h"
 #include "cesium/omniverse/Tokens.h"
 #include "cesium/omniverse/UsdUtil.h"
 
-#ifdef CESIUM_OMNI_MSVC
-#pragma push_macro("OPAQUE")
-#undef OPAQUE
-#endif
-
-#include <CesiumGltf/Model.h>
 #include <omni/fabric/FabricUSD.h>
-#include <omni/fabric/IFabric.h>
-#include <omni/ui/ImageProvider/DynamicTextureProvider.h>
 #include <spdlog/fmt/fmt.h>
-
-#include <array>
 
 namespace cesium::omniverse {
 
 FabricMaterial::FabricMaterial(
-    pxr::SdfPath path,
+    const omni::fabric::Path& path,
     const FabricMaterialDefinition& materialDefinition,
-    pxr::SdfAssetPath defaultTextureAssetPath,
+    const pxr::TfToken& defaultTextureAssetPathToken,
     long stageId)
-    : _materialDefinition(materialDefinition)
-    , _defaultTextureAssetPath(std::move(defaultTextureAssetPath))
+    : _materialPath(path)
+    , _materialDefinition(materialDefinition)
+    , _defaultTextureAssetPathToken(defaultTextureAssetPathToken)
     , _stageId(stageId) {
+
     if (stageDestroyed()) {
         return;
     }
 
-    initialize(std::move(path), materialDefinition);
+    initialize();
     reset();
 }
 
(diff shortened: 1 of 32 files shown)


src/core/src/FabricMaterial.cpp
```diff
@@ -310,18 +303,16 @@ void FabricMaterial::setBaseColorTextureValues(
...
+    auto textureFabric = srw.getAttributeWr<omni::fabric::AssetPath>(texturePath, FabricTokens::inputs_texture);
+    auto texCoordIndexFabric = srw.getAttributeWr<int>(texturePath, FabricTokens::inputs_tex_coord_index);
+    auto wrapSFabric = srw.getAttributeWr<int>(texturePath, FabricTokens::inputs_wrap_s);
+    auto wrapTFabric = srw.getAttributeWr<int>(texturePath, FabricTokens::inputs_wrap_t);
+    auto offsetFabric = srw.getAttributeWr<pxr::GfVec2f>(texturePath, FabricTokens::inputs_offset);
+    auto rotationFabric = srw.getAttributeWr<float>(texturePath, FabricTokens::inputs_rotation);
+    auto scaleFabric = srw.getAttributeWr<pxr::GfVec2f>(texturePath, FabricTokens::inputs_scale);
 
-    textureFabric->assetPath = pxr::TfToken(textureAssetPath.GetAssetPath());
-    textureFabric->resolvedPath = pxr::TfToken(textureAssetPath.GetResolvedPath());
```
lilleyse, line 324: In a couple places I was creating these temporary `TfToken`'s which is unsafe since tokens are reference counted and destroyed when the count reaches zero.

Now the texture asset path token is stored in the `FabricTexture` so the token stays around as long as the `FabricTexture` exists.

That's also why the `cesium.mdl` path is passed along as a token instead of an `SdfAssetPath`. The token will exist as long as `Context` exists.

src/core/src/FabricMaterial.cpp
```diff
@@ -62,175 +52,177 @@ void FabricMaterial::setActive(bool active) {
...
+    *inputsExcludeFromWhiteModeFabric = false;
+    *infoImplementationSourceFabric = FabricTokens::sourceAsset;
+    infoMdlSourceAssetFabric->assetPath = Context::instance().getCesiumMdlPathToken();
+    infoMdlSourceAssetFabric->resolvedPath = pxr::TfToken();
+    *infoMdlSourceAssetSubIdentifierFabric = FabricTokens::cesium_material;
+
+    if (hasVertexColors) {
+        const auto vertexColorPrimvarNameSize = UsdTokens::vertexColor.GetString().size();
+        srw.setArrayAttributeSize(shaderPath, FabricTokens::inputs_vertex_color_name, vertexColorPrimvarNameSize);
+        auto vertexColorNameFabric =
```
corybarr, line 147: What's the use case for a vertex color name?

corybarr (approved):
This is my first exposure to the Connection API so I don't understand the code fully, but the approach and code makes sense.

src/core/src/OmniTileset.cpp
```diff
@@ -280,6 +280,10 @@ TilesetStatistics OmniTileset::getStatistics() const {
 }
 
 void OmniTileset::reload() {
+    if (_renderResourcesPreparer != nullptr) {
+        _renderResourcesPreparer->detachTileset();
+    }
```
lilleyse, line 285: This was something that I missed in https://github.com/CesiumGS/cesium-omniverse/pull/416.

If the cesium-native tileset is destroyed (like when a tileset reload occurs) we need to tell the render resource preparer to detach its tileset reference.

src/core/src/UsdNotificationHandler.cpp
```diff
@@ -89,6 +89,16 @@ void UsdNotificationHandler::onObjectsChanged(const pxr::UsdNotice::ObjectsChang
...
             if (UsdUtil::primExists(path)) {
+                const auto isTileset = getType(path) == ChangedPrimType::CESIUM_TILESET;
+                const auto alreadyRegistered = AssetRegistry::getInstance().getTilesetByPath(path).has_value();
+
+                if (isTileset && alreadyRegistered) {
+                    // A prim may be resynced even if its path doesn't change, like when an API Schema is applied to it
+                    // This happens when a material is assigned to a tileset for the first time
+                    // We don't want to add the prim again if it's already registered
+                    continue;
+                }
```
lilleyse, line 100: See https://github.com/CesiumGS/cesium-omniverse/pull/406#issuecomment-1663109737

lilleyse (commented):
Some inline comments to help the review

src/core/src/FabricMaterial.cpp
lilleyse: This is the fabric attribute that populates the `vertex_color_name` uniform in `gltf/pbr.mdl`, since we want to override the default vertex color name.

See `extern/nvidia/_build/target-deps/kit-sdk/mdl/core/mdl/gltf/pbr.mdl`

lilleyse:
> This is my first exposure to the Connection API so I don't understand the code fully, but the approach and code makes sense.

We started using the Connection API in Kit 105 and it simplifies how inputs and outputs are connected for MDL. Previously we had to manage the terminals and relationships ourselves. You can see some of the changes that took place in https://github.com/CesiumGS/cesium-omniverse/pull/382.

Issue #310, opened 2023-05-31
Explore how to run tests against the USD stage


weegeekps:
We have various needs for doing simple testing against just a USD stage. We need to explore how to do this with Omniverse and then document and share it with the rest of the team.

lilleyse:
Most of the C++ code depends on `Context` so we should start there. The testing framework will need to initialize and manage the `Context` as if it were a Kit application - e.g. calling `onStartup` and `onShutdown`, passing in a USD stage, and updating the frame.

In theory it should be simple for the testing code to create a USD stage and pass it to `Context::setStageId`, but it may be more complicated than that if there's any Omniverse specific behavior we need to replicate. Also we'll need to figure out how to assign a Fabric `StageReaderWriter` to the USD stage.

mattelser:
@lilleyse suggested tests for the usd utils functions [in this PR](https://github.com/CesiumGS/cesium-omniverse/pull/428#discussion_r1296009801_), which could be a good place to start

PR #428, opened 2023-08-15
remove thin USD wrappers from C++ API in favor of underlying USD functions

### LINKED ISSUE #390: Remove functions from C++ API

mattelser:
fixes #390

The only thing that calls remove_tileset is clicking the `remove all tilesets` button of the debug window


lilleyse:
I'm still going through my review, but as an early comment these two areas need to be updated:

* https://github.com/CesiumGS/cesium-omniverse/blob/252d8cf2c017fcd30245ea370d3da477837b3d54/exts/cesium.omniverse/cesium/omniverse/extension.py#L313-L315
* https://github.com/CesiumGS/cesium-omniverse/blob/993e8f2bb983a991994908b0c6eaa7e977c0967f/exts/cesium.omniverse/cesium/omniverse/ui/asset_details_widget.py#L78

exts/cesium.omniverse/cesium/omniverse/utils/utils.py
lilleyse, line 1: Style notes throughout the code:

* use camel_case for variable names
* add return type annotations

exts/cesium.omniverse/cesium/omniverse/utils/utils.py
```diff
@@ -34,3 +42,50 @@ def str_is_empty_or_none(s: Optional[str]):
...
+    parent = imagery.GetPrim().GetParent()
+    assert parent.IsA(CesiumTileset)
+
+    imagery.GetIonAssetIdAttr().Set(asset_id)
+    imagery.GetIonAccessTokenAttr().Set(token)
+
+    return imagery_path
+
+
+def is_tileset(maybeTileset):
```
lilleyse, line 84: `maybeTileset` should have a type annotation.

exts/cesium.omniverse/cesium/omniverse/utils/utils.py
lilleyse, line 1: For consistency and simplicity, I think both arguments and return variables for the USD util functions should be `str` types.

lilleyse, line 1: Could you write tests for the new functions in here?

lilleyse:
Other things:

* CI is failing
* Python unit tests are failing (though this may be happening on `main` as well)
* Could you merge `main` into this branch?

exts/cesium.omniverse/cesium/omniverse/utils/utils.py
```diff
@@ -34,3 +42,50 @@ def str_is_empty_or_none(s: Optional[str]):
         return True
 
     return False
+
+
+def add_tileset_ion(name: str, tilesetId: int, token: str = ""):
```
weegeekps, line 47: None of the tileset stuff belongs in this file. It should be moved to a different file. The utils file is for small UI related utilities.

exts/cesium.omniverse/cesium/omniverse/utils/utils.py
mattelser: agreed those would be a good idea, but probably out of scope for this PR. I've suggested them in #310

exts/cesium.omniverse/cesium/omniverse/utils/utils.py
mattelser: fair point, I've moved these into a `usdUtils` module

exts/cesium.usd.plugins/cesium/usd/plugins/CesiumUsdSchemas/__init__.pyi
```diff
@@ -146,7 +146,7 @@ class Session(pxr.Usd.Typed):
     @classmethod
     def __reduce__(cls) -> Any: ...
 
-class Tileset(pxr.UsdGeom.Boundable):
+class Tileset(pxr.UsdGeom.Gprim):
```
lilleyse, line 149: Thanks for fixing this 👍

lilleyse (approved):

exts/cesium.omniverse/cesium/omniverse/utils/utils.py
weegeekps: Thanks for fixing this.

weegeekps (approved):

PR #438, opened 2023-08-16
Simplify token creation

lilleyse:
The previous approach for adding new tokens was overly complex, requiring changes in 5 different places within `Tokens.h` and `Tokens.cpp`. Now it's as simple as adding a new token to `USD_TOKENS` and letting the macros do the rest of the work.

USD tokens are now created in a more standard way with `TF_DECLARE_PUBLIC_TOKENS` and `TF_DEFINE_PUBLIC_TOKENS` and are accessed as `pxr::UsdTokens->tokenName`.

Fabric tokens are created with `FABRIC_DECLARE_TOKENS` and `FABRIC_DEFINE_TOKENS` which are inspired by the pxr macros and are accessed as `FabricTokens::tokenName` like before.

### DIFF
--- a/src/core/include/cesium/omniverse/Tokens.h
+++ b/src/core/include/cesium/omniverse/Tokens.h
@@ -1,136 +1,107 @@
 #pragma once
 
-#include <omni/fabric/SimStageWithHistory.h>
-#include <pxr/base/tf/token.h>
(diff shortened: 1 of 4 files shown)


weegeekps (approved):
Lots cleaner. :shipit:

Issue #425, opened 2023-08-15
Unhandled exception when creating a tileset for the second time


corybarr:
![image](https://github.com/CesiumGS/cesium-omniverse/assets/1040194/dc76b70a-f7a5-4212-9d37-b85269cd49ef)

To reproduce:

1. Load a tileset
2. Delete the tileset
3. Reload the same tileset

lilleyse:
Interestingly, the crash only happens if the tileset has materials. If you check "Disable Materials" on the Cesium prim and follow the steps it doesn't crash.

corybarr:
I'm also able to reproduce this without creating a new tileset. I've attached a .usd file to reproduce. These steps should trigger the issue. (You probably need to be on the `add-voxel-point-cloud` branch.)

1. Delete the Melbourne_Point_Cloud prim
2. Enable visibility on the Melbourne_Photogrammetry prim
3. The exception should trigger.

![image](https://github.com/CesiumGS/cesium-omniverse/assets/1040194/d0308258-61a5-4fa6-be38-dac25f97c2aa)

[melbourne.zip](https://github.com/CesiumGS/cesium-omniverse/files/12352098/melbourne.zip)

lilleyse:
I'm starting to narrow this down some more. If we retain a reference to the deleted prim's path forever it doesn't crash. That means something in Omniverse is referencing the deleted prim's path well after it's been destroyed. It's strange that this is only a problem for materials.

With my hacky fix I no longer see any crashes when repeatedly reloading CWT. I still see crashes with CWT + Bing but that's a different crash related to Hydra and textures. I'm still narrowing in on that one.

lilleyse:
@mattelser pointed out that if if you click "remove all tilesets" in the debugging window and then create the tileset again it'll hang instead of crashing. I can reproduce that.

lilleyse:
Note also the error printed right before the crash:

> 2023-08-17 21:46:51 [358,281ms] [Warning] [omni.usd] Coding Error: in _PossiblyDestroyRep at line 168 of W:\ac88d7d902b57417\USD\pxr\base\tf\token.cpp -- Failed verification: ' repFoundInSet ' -- failed to find token '' in table for destruction

Which corresponds to [this line in USD](https://github.com/PixarAnimationStudios/OpenUSD/blob/10b62439e9242a55101cf8b200f2c7e02420e1b0/pxr/base/tf/token.cpp#L166-L168).

Here's the call stack:

[...]

PR #439, opened 2023-08-16
Clean up Fabric token and path usage

lilleyse:
I'm always a bit weary of how we're handling tokens and paths because they're reference counted and can lead to hard to fix bugs if handled incorrectly. So I swept through the code and tried to standardize on some better practices for dealing with tokens and paths.

* Changed `getAttribute` calls so that they return `TokenC` instead of `Token`. This makes token lifetime a bit clearer. There's no need to use the reference counted type when Fabric is just going to memcpy the value. Technically we could use `uint64_t` everywhere but it's clearer if we use `TokenC` which is a simple struct wrapper around a `uint64_t`. Same reason for switching to `PatchC` instead of `Path` whenever we're getting or setting Fabric data.
* In all other places, let Fabric do implicit conversions between `Token`/`TokenC` and `Path`/`PathC` instead of doing it ourselves
* Return const references to paths in a few areas

There will probably be some follow ups as I investigate https://github.com/CesiumGS/cesium-omniverse/issues/425

### DIFF
--- a/src/core/include/cesium/omniverse/FabricAttributesBuilder.h
+++ b/src/core/include/cesium/omniverse/FabricAttributesBuilder.h
@@ -6,7 +6,7 @@ namespace cesium::omniverse {
 
 class FabricAttributesBuilder {
   public:
-    void addAttribute(const omni::fabric::Type& type, const omni::fabric::TokenC& name);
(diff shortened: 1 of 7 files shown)


weegeekps (approved):
LGTM

PR #440, opened 2023-08-17
Fix USD Composer symlink

lilleyse:
https://github.com/CesiumGS/cesium-omniverse/pull/426/ added the material graph extension to our development kit app. Since material graph is only included in certain apps like USD Composer and not kit-sdk we have to locate USD Composer and create a symlink to it. The CMake code to create the symlink was broken.

### DIFF
--- a/extern/CMakeLists.txt
+++ b/extern/CMakeLists.txt
@@ -37,7 +37,7 @@ add_external_project(
 # cmake-format: on
 
 if(UNIX)
-    execute_process(COMMAND bash "${PROJECT_SOURCE_DIR}/extern/nvidia/prebuild.sh" RESULT_VARIABLE exit_code)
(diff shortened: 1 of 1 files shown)


extern/CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ endif()
 # Add a symlink to USD Composer (create) so that we can use its extensions (e.g. omni.kit.window.material_graph) in our internal applications
 # Don't check for errors because we can still proceed with the build
```
weegeekps, line 52: We should check for errors, because while the build may still pass if this fails, we can't run the test app.

weegeekps: Can we at least log a clear warning to the build output then? This feels like a major footgun.

lilleyse: Yeah I can do that

lilleyse: Done

image

weegeekps (approved):
Works good. A side effect of this is that we do have to have the OV launcher running in the background again, but that's not a terrible thing.

PR #442, opened 2023-08-18
Fix crash when reloading tileset

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/425

This is a workaround for https://github.com/CesiumGS/cesium-omniverse/issues/425. By retaining a reference to the path we avoid a crash deep in the Sdf library that's triggered by loading a tileset, removing the tileset, and reloading the tileset. It's possible this will be fixed in a future Kit release at which point we can remove this workaround.

I'm tracking this in https://github.com/CesiumGS/cesium-omniverse/issues/243


lilleyse:
The best way to test this is to load Cesium World Terrain (untextured) from the Add Asset window.

Textures are causing different crashes that I'm still looking into.

weegeekps (approved):
Works fine.

PR #447, opened 2023-08-18
increase window wait for ui tests until window doesn't require delay

### LINKED ISSUE #437: python test failures
Python tests currently fail. There look to be two problems involved:

- `69833c4f32375488ab80c1114ff0ba41c9bc74cd` introduced a delay in spawning the cesium window and fabric modal. I think the tests are expecting these windows to have spawned before this delay is done because reverting this commit causes these tests to pass
- the `pass_fail_widget_failed` test fails inconsistently. This is potentially consistently inconsistent? So far, if I reset to the current tip of the main branch then revert the above commit, this test will fail on the first run then succeed on subsequent runs. This is very weird and I'll look into it further

mattelser:
fixes #437 . The delay introduced in spawning the cesium window in #423 caused UI tests to run before the windows had actually spawned. Adding a timeout to the tests with a comment for context

### DIFF
--- a/exts/cesium.omniverse/cesium/omniverse/tests/extension_test.py
+++ b/exts/cesium.omniverse/cesium/omniverse/tests/extension_test.py
@@ -13,6 +13,12 @@ _window_ref: Optional[ui_test.WidgetRef] = None
 class ExtensionTest(omni.kit.test.AsyncTestCase):
     async def setUp(self):
         global _window_ref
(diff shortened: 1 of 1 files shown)


lilleyse:
I can confirm the tests pass on Windows

weegeekps (approved):
Works great on Linux.

Issue #450, opened 2023-08-22
Performance app crashes


mattelser:
The performance app may have multiple issues. 
- Currently it crashes before opening, though @lilleyse found a fix for that (PR pending). 
- Once that is in place and the app opens, however, clicking a first location works but then the app crashes when a second location is selected. This may be an extension of #425, though notably still happens after that workaround was merged.

lilleyse:
I don't see the crashes anymore, most likely fixed by https://github.com/CesiumGS/cesium-omniverse/pull/455.

Though sometimes the tilesets don't load at all when switching between locations...

mattelser:
I also don't see crashes anymore. Closing.

PR #451, opened 2023-08-22
partial fix for performance test crashes

mattelser:
partial fix for #450. @lilleyse found that this allows the performace app to load without crashing. CI appears to be failing for server-side issues, I'll retry those jobs shortly

### DIFF
--- a/apps/cesium.performance.kit
+++ b/apps/cesium.performance.kit
@@ -9,3 +9,11 @@ app = true
 
 [settings]
 app.window.title = "Cesium For Omniverse Performance Testing App"
+
+[settings.app.exts]
+folders.'++' = [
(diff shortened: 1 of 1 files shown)


mattelser:
CI related issues have cleared, and this passes now

lilleyse:
Thanks @mattelser.

Just to summarize why the fix works: the important line is `"${app}/exts"`. Otherwise `cesium.performance.kit` can't find `cesium.performance.app` which is in the `exts` subfolder. This doesn't affect the other `.kit` applications because their dependencies are all in the same folder.


Unfortunately since we're overriding `[settings.app.exts]` we need to include all the paths needed by dependencies.

lilleyse (approved):

PR #454, opened 2023-08-24
Fix USD Composer symlink

lilleyse:
@corybarr and I both had trouble symlinking to USD Composer in Windows, possibly due to IPv4 vs. IPv6 configuration on our machines.

Fixed by using `localhost` instead of `127.0.0.1` in `link_app.py`.


corybarr (approved):
Looks good, makes sense, very simple explanation

weegeekps (approved):
Works fine on Linux.

PR #456, opened 2023-08-28
Add CUDA to build system

lilleyse:
Adds cuda and nvrtc to the build system in preparation for @corybarr's point cloud PRs.

A couple things to note:

* We're now installing nvidia drivers on Linux CI so that we can link against libcuda.so. The reason we don't need to do this on Windows is because we can link against the import library nvcuda.lib which is provided in packman. As far as I can tell there's no equivalent workflow in Linux.
* nvrtc doesn't ship with kit so we need to install it manually in `exts/cesium.omniverse/bin` and include it in our release zip. Unfortunately this adds an extra 40MB to the zip. It would be better if we did kernel compilation at build time instead of runtime, then we could drop nvrtc. @corybarr is this possible?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -109,7 +109,7 @@ jobs:
         if: ${{ runner.os == 'Linux' && matrix.config.name != 'CentOS 7 - GCC' }}
         run: |
           sudo apt update
-          sudo apt install -y doxygen clang-tidy-14 gcovr
+          sudo apt install -y doxygen clang-tidy-14 gcovr nvidia-driver-535
           sudo pip3 install conan==1.58.0 black==23.1.0 flake8==6.0.0
 
       # Could not figure out how to add Chocolatey libraries to the PATH automatically with refreshenv, so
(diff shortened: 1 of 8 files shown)


corybarr (approved):
Looks good, builds for me

lilleyse:
> If we end up doing something like shader variants where we compile on-the-fly based on something like styling choices, then we'd need nvrtc or would have to compile the variants.

Yeah I was thinking the same. It could be really powerful for on-the-fly styling. Though anything that can be built up-front should be built up-front.

lilleyse:
Let's hold off on merging this until after the release. No point in adding 40MB into the release zip until we're actually using nvrtc. I'll mark as draft until then.

lilleyse:
https://github.com/CesiumGS/cesium-omniverse/pull/456/commits/671489ad19f9bb8055fd52eec41dfc83e89652c7 fixed CentOS builds. This should be ready to go now.

corybarr:
@lilleyse I noticed a runtime error if `nvrtc-builtins64_118.dll` wasn't also added. Might make sense to add it to another PR.

lilleyse:
@corybarr follow up PR for nvrtc-builtins: https://github.com/CesiumGS/cesium-omniverse/pull/465

PR #458, opened 2023-08-28
provide a more helpful error message when failing to fetch a tileset

### LINKED ISSUE #321: Bad error message when tileset URL doesn't exist
When setting the tileset URL to an invalid path like `file:///C:/fake/path` we got a pretty uninformative error message

```
2023-06-07 22:43:05  [Error] [cesium.omniverse.plugin] [2023-06-07 18:43:05.703] [cesium-omniverse] [error] Error when parsing tileset JSON, error code 1 at byte offset 0
2023-06-07 22:43:05  [Error] [cesium.omniverse.plugin] 
```

mattelser:
fixes #321. I tested with both a bad and a good file path, any further suggestions welcome

### DIFF
--- a/src/core/src/HttpAssetAccessor.cpp
+++ b/src/core/src/HttpAssetAccessor.cpp
@@ -4,6 +4,8 @@
 
 #include <omni/kit/IApp.h>
 #include <zlib.h>
+#include <exception>
+#include <stdexcept>
 
 namespace cesium::omniverse {
 namespace {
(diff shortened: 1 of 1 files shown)


src/core/src/HttpAssetAccessor.cpp
```diff
@@ -109,7 +112,11 @@ CesiumAsync::Future<std::shared_ptr<CesiumAsync::IAssetRequest>> HttpAssetAccess
     session->SetHeader(cprHeader);
     session->SetUrl(cpr::Url(url));
     session->GetCallback([promise, url, headers](cpr::Response&& response) mutable {
-        promise.resolve(std::make_shared<HttpAssetRequest>("GET", url, headers, std::move(response)));
+        if (response.error) {
+            promise.reject(std::runtime_error("failed to fetch tileset from: " + url + "\n"));
```
lilleyse, line 116: Style thoughts:

* Capitalize the error message
* Use fmt for string formatting rather than string concatenation
* Is the newline needed?

Also, it might make sense to include the response error in the message.

lilleyse: Also, this error can get printed for any types of requests, not just tilesets, so the message should be more generic, like what cesium-unity is doing: https://github.com/CesiumGS/cesium-unity/blob/82c3be4f7cfd0b3f120e57b75771f5d3b46c222d/native~/Shared/src/UnityAssetAccessor.cpp#L187-L188

mattelser: Thanks! Incorporated into latest commit

lilleyse (approved):
Thanks @mattelser, just merge main into this branch and it should be good to go.

https://github.com/CesiumGS/cesium-omniverse/pull/457 also updated CHANGES.md so watch out for merge conflicts.

PR #462, opened 2023-08-31
Reuse materials where possible

### LINKED ISSUE #344: Reuse materials where possible
In general, it's not possible for prims to share materials since material values and texture paths are baked into the material definition, except in limited cases where primvars are used.

However, if materials are truly identical (same PBR values, untextured) we should be able to share the same material. This would mean updating `ObjectPool` so that it does reference counting for active materials and only releases materials once the count reaches zero.

An example where this would be helpful is when loading Cesium World Terrain without imagery.

lilleyse:
Fixes #344 so that tiles with identical materials now reference the same FabricMaterial instead of allocating an object pool. This is for tilesets with untextured materials like CWT (without imagery), Cesium OSM Buildings, and point clouds. It should be a pretty good performance improvement for these types of tilesets.

Most of the code changes are in FabricResourceManager. A SharedMaterial object encapsulates a FabricMaterial, its material values, and a reference count. When calling `FabricResourceManager::acquireMaterial` it checks if there's an existing material with the same material values and returns that. When the material is released the reference count is decremented and if it hits zero the material is destroyed.

You can test this by adding Cesium World Terrain without imagery and seeing that only one material is being used.

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,12 +4,16 @@
 
 ### v0.10.0 - 2023-09-01
 
-* Added basic point cloud support
-* Fixed loading extension in Omniverse Code 2023.1.1
-* Fixed crashes when reloading tileset
-* Fixed memory leak when removing tileset mid-load
-* Fixed several other bugs related to removing tilesets mid-load
-* Upgraded to cesium-native v0.26.0
+* Reduced the number of materials created when loading untextured tilesets.
+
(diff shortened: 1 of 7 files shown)


corybarr (approved):
Looks clean. Makes sense.

lilleyse:
Almost forgot to update `CHANGES.md`. That's done now.

PR #468, opened 2023-09-13
Set up basic tests extension to test c++ code

mattelser:
While I am working on the actual tests to be run in this extension, it probably makes sense to merge this into main. I had hoped to simplify the python bindings just to be a basic class, but either Carbonite requires there to be a child of its interface class, or I was doing something wrong. @lilleyse and I got it working by inheriting from an interface class, so for the sake of time we can proceed with that.


.vscode/launch.linux.json
```diff
@@ -29,6 +29,34 @@
...
+      "setupCommands": [
+        {
+          "text": "-enable-pretty-printing",
+          "ignoreFailures": true
+        },
+        {
+          "text": "set print elements 0"
+        }
+      ]
+    },
```
lilleyse, line 59: Add an equivalent configuration to `launch.windows.json`

.gitignore
```diff
@@ -46,10 +46,13 @@ exts/cesium.omniverse/cesium/omniverse/bindings/CesiumOmniversePythonBindings.cp
 exts/cesium.usd.plugins/cesium/usd/plugins/CesiumUsdSchemas/_CesiumUsdSchemas.so
 exts/cesium.usd.plugins/cesium/usd/plugins/CesiumUsdSchemas/_CesiumUsdSchemas.pyd
 exts/cesium.usd.plugins/cesium/usd/plugins/CesiumUsdSchemas/_CesiumUsdSchemas.lib
+exts/cesium.tests/cesium/tests/bindings/CesiumOmniverseTestsPythonBindings.cpython-310-x86_64-linux-gnu.so
+exts/cesium.tests/cesium/tests/bindings/CesiumOmniverseTestsPythonBindings.cp310-win_amd64.pyd
 
 # Installed libraries
 exts/cesium.omniverse/bin/
 exts/cesium.usd.plugins/bin/
+exts/cesium.tests/bin/
```
lilleyse, line 55: I think the kit extension should be renamed to `cesium.omniverse.tests` so that it's clear that these are tests for the `cesium.omniverse` extension.

CMakeLists.txt
```diff
@@ -437,6 +437,7 @@ add_subdirectory(src)
 if(CESIUM_OMNI_ENABLE_TESTS)
     enable_testing()
     add_subdirectory(tests)
+    add_subdirectory(cesiumTestsExtension)
```
lilleyse, line 440: It's ok that the folder is called `cesiumTestsExtension` in this PR, but once we migrate the existing tests we should rename the folder back to `tests`.

CMakeLists.txt
lilleyse, line 1: Do you know what it might take to get coverage working with this new test extension setup?

lilleyse, line 1: Could you add the folders in `cesiumTestsExtension` to `LINT_SOURCE_DIRECTORIES`? That way CI will catch any formatting errors.

cesiumTestsExtension/bindings/PythonBindings.cpp
```diff
@@ -0,0 +1,20 @@
+#include <carb/BindingsPythonUtils.h>
+#include "CesiumTests.h"
```
lilleyse, line 2: Looks like this file isn't formatted correctly. Could you run the formatter? (In VSCode, Ctrl + Shift + B then Format)

Note you will need to update `LINT_SOURCE_DIRECTORIES` for this work.

cesiumTestsExtension/bindings/PythonBindings.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+// NOLINTNEXTLINE
+CARB_BINDINGS("cesium.tests.python")
```
lilleyse, line 5: Similar to the other comment, rename this to `cesium.omniverse.tests.python`

cesiumTestsExtension/bindings/pyboost11.h
lilleyse, line 1: I think you can safely remove pyboost11.h from here since we don't need to do any of the pyboost -> pybind11 binding magic that the main extension is doing.

cesiumTestsExtension/include/CesiumTests.h
lilleyse, line 1: Similar to other comments, rename to CesiumOmniverseTests.

cesiumTestsExtension/bindings/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+setup_python_module(
+    PYTHON_DIR
+        # Use the same Python version as Omniverse (Python 3.10)
+        "${PROJECT_SOURCE_DIR}/extern/nvidia/_build/target-deps/python"
+    TARGET_NAME
+        CesiumOmniverseTestsPythonBindings
+    SOURCES
+        ${SOURCES}
+    LIBRARIES
+        doctest::doctest
```
lilleyse, line 15: You should be able to remove `doctest::doctest` from here since `doctest::doctest` is a dependency `cesium.omniverse.tests.plugin` and our CMake setup handles the library dependency chain.

cesiumTestsExtension/public/CesiumTests.cpp
```diff
@@ -0,0 +1,93 @@
...
+        Context::onStartup(cesiumExtensionLocation);
+    }
+
+    void onShutdown() noexcept override {
+        Context::onShutdown();
+    }
+
+    void run_all_tests(long int stage_id) noexcept override {
+
+        std::cout << "Running Cesium Omniverse Tests with stage id: " << stage_id << std::endl;
```
lilleyse, line 65: Use `CESIUM_LOG_INFO` or one of the other logging macros instead of `std::cout`.

cesiumTestsExtension/public/CesiumTests.cpp
```diff
@@ -0,0 +1,93 @@
...
+        if (ad.m_threw)
+            std::cout << ad.m_exception;
+        else
+            std::cout << "  values: " << assertString(ad.m_at) << "( " << ad.m_decomp << " )";
+    } else {
+        std::cout << Color::None << "an assert dealing with exceptions has failed!";
+    }
+
+    std::cout << std::endl;
+}
```
lilleyse, line 49: >    // This file is set up such that we can use doctest assertions (e.x. CHECK)
>    // outside of a test suite.

Is there a reason that we're running doctests assertions outside of a test suite? It would be nice to avoid all this assertion handling.

exts/cesium.tests/cesium/tests/test_window.py
lilleyse, line 1: Was this file accidentally checked in?

exts/cesium.tests/cesium/tests/utils/__init__.py
lilleyse, line 1: Same comment

exts/cesium.tests/cesium/tests/utils/utils.py
lilleyse, line 1: Same comment

exts/cesium.tests/config/extension.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+version = "0.1.0"
+category = "simulation"
+feature = false
+app = false
+title = "Cesium for Omniverse Test"
```
lilleyse, line 6:
```suggestion
title = "Cesium for Omniverse Tests"
```

exts/cesium.tests/config/extension.toml
```diff
@@ -0,0 +1,33 @@
...
+
+# Paths are relative to the extension folder
+readme = "doc/README.md"
+icon = "doc/resources/icon.png"
+
+[[python.module]]
+name = "cesium.tests"
+
+[python.pipapi]
+archiveDirs = ["vendor"]
```
lilleyse, line 33: This can likely be removed since this extension itself doesn't import anything from pip.

exts/cesium.tests/doc/README.md
```diff
@@ -0,0 +1,13 @@
+# Cesium for Omniverse Test Extension
```
lilleyse, line 1:
```suggestion
# Cesium for Omniverse Tests Extension
```

exts/cesium.tests/doc/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+This extension is designed to run tests against the Cesium for Omniverse Extension.
+
+
+## Current Features
+
+### TODO list features
```
lilleyse, line 8: I would just remove these headers

exts/cesium.tests/doc/resources/icon.png
lilleyse, line 1: Is there a path issue somewhere? Because neither the README nor the icon show up in the extensions window.

image

.gitignore
lilleyse: Note: as part of this you will need to rename `cesium.omniverse.tests.kit` to something else otherwise the name conflict will cause issues. Maybe `cesium.omniverse.tests.runner.kit`.

mattelser: I think the omniverse component of the name is unnecessary and deviates from the convention used in our other extensions. These extensions only exist in a context where the relationship to omniverse is clear: in the cesium-omniverse repo or (if deployed) in Omniverse itself.

cesiumTestsExtension/public/CesiumTests.cpp
mattelser: I tried defining a test suite/case previously, but couldn't make it work. Doctest does provide the ability to define test cases within classes (via the TEST_CASE_CLASS macro), but it seems to require instantiating the object, which for the interface class is tricky if it's doable at all. I couldn't even get a simpler test case to compile in these classes though. If I placed a test case inside a method, the compiler bans a function definition that the macro does. If I placed it outside a method and in the header as the example doc shows, I get a compilation error that suggests an internal doctest issue (`error: ‘DOCTEST_ANON_VAR_15’ was not declared in this scope`)

Thankfully the one handler can cover all the tests

.gitignore
lilleyse: I'm not sure if this is future proof though. We may want to publish other extensions and write tests for those, in which case it's not clear what `cesium.tests` is testing.

I could imagine eventually

`cesium.powertools` -> `cesium.powertools.tests`
`cesium.usd.plugins` -> `cesium.usd.plugins.tests`
`cesium.extensionInOtherRepo` -> `cesium.extensionInOtherRepo.tests`

There's already some precedence for the `.tests` suffix

image

This change would also make the extension name and carbonite plugin name more consistent, as we call the carbonite plugin `cesium.omniverse.tests.plugin` and the library `CesiumOmniverseTests`.

lilleyse: That said, it's pretty rare to see kit extensions have a separate tests extension except the few examples in the screenshot above above. So maybe we need to reevaluate this at some point.

cesiumTestsExtension/public/CesiumTests.cpp
lilleyse: Sort of a follow up question: is it possible to see a test report similar to what we get with the `tests` executable?

```
[doctest] doctest version is "2.4.9"
[doctest] run with "--help" for options
===============================================================================
[doctest] test cases:   10 |   10 passed | 0 failed | 0 skipped
[doctest] assertions: 1277 | 1277 passed | 0 failed |
[doctest] Status: SUCCESS!
```

Currently, we just see which tests have failed. And without `TEST_SUITE` there's no grouping of tests.

It doesn't necessarily need to hold up the PR, but eventually I would hope for something a bit more ergonomic.

mattelser: agreed

CMakeLists.txt
mattelser: I hope to rig up something, but that sounds like a ticket of its own

cesiumTestsExtension/public/CesiumTests.cpp
mattelser: I've just tried using CESIUM_LOG_INFO, but it turns out `onStartup` isn't actually being called so context isn't initialized, which causes the logger to segfault. It's just a matter of adding the bindings for `onStartup` (and `onShutdown`) then passing the extension location down to them, which I wasn't able to get done today.

mattelser:
Just pushed a version that addresses most of the review comments, with some discussions unresolved. I'm gone next week but can sort out the remainder once I get back.

.gitignore
mattelser: Done. I needed to change the name to `cesium.omniverse.cpp.tests` to avoid a conflict with the python module in `cesium.omniverse/tests`

corybarr (approved):
This isn't an area I'm an expert in, but this all seems good to me. Seems straightforward to drop in a new test.

PR #472, opened 2023-09-28
fix build issue in release, tests were trying to install but not build

mattelser:
There was no conditional preventing tests from trying to be installed when building in release mode, however they were not being built leading to the error. I don't think we want these tests in release so I've prevented the install rather than enabling a build.


corybarr (approved):
Makes sense, looks good

Issue #266, opened 2023-04-27
Add support for compressed textures and mipmaps


lilleyse:
`DynamicTextureProvider` supports a bunch of different image formats included compressed formats.

```
(in carb_sdk_plugins/include/carb/RenderingTypes.h)

enum class Format
{
    ...
    eBC1_RGBA_UNORM,
    eBC1_RGBA_SRGB,
    eBC2_RGBA_UNORM,
    eBC2_RGBA_SRGB,
    eBC3_RGBA_UNORM,
    eBC3_RGBA_SRGB,
    eBC4_R_UNORM,
    eBC4_R_SNORM,
    eBC5_RG_UNORM,
    eBC5_RG_SNORM,
    eBC6H_RGB_UFLOAT,
    eBC6H_RGB_SFLOAT,
    eBC7_RGBA_UNORM,
    eBC7_RGBA_SRGB,
    ...
};
```

It also supports setting mipmap data. See `setMipMappedBytesData` (both overloads) in `kit-sdk/dev/include/omni/ui/ImageProvider/ByteImageProvider.h`.

It should be pretty straightforward to support KTX2/Basis textures coming from cesium-native this way. Also see https://github.com/CesiumGS/cesium-unity/issues/274.

lilleyse:
For testing see:

[agi-ktx2-explicit.zip](https://github.com/CesiumGS/cesium-omniverse/files/11409647/agi-ktx2-explicit.zip)
[agi.usda.txt](https://github.com/CesiumGS/cesium-omniverse/files/11409650/agi.usda.txt)

PR #474, opened 2023-10-02
Decode BasisU textures to BCn format instead of RGBA8

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/266

Previously, compressed textures were decoded to RGBA8 instead of the native BCn format due to a limitation in Kit 105.0. Now with Kit 105.1 we can decode to BCn and keep the data compressed on the GPU. This involves a bit of a strange workaround with the `stride` parameter (see comments in the code).

Test data:

[agi-ktx2-explicit.zip](https://github.com/CesiumGS/cesium-omniverse/files/11409647/agi-ktx2-explicit.zip)
[agi-ktx.usda.zip](https://github.com/CesiumGS/cesium-omniverse/files/11972717/agi-ktx.usda.zip)


lilleyse:
I confirmed that the texture memory is lower because of this change. See the `Texture - Other` category. It's a small difference (112 MB - 101 MB), but I think it's because there's other textures being counted in that stat and the initial view doesn't load many tiles. I ran multiple runs and get the same result.

Before|After
--|--
before|after

src/core/src/Context.cpp
```diff
@@ -637,6 +641,15 @@ const pxr::TfToken& Context::getCesiumMdlPathToken() const {
     return _cesiumMdlPathToken;
 }
 
+bool Context::isCompressedTexturesSupported() const {
+    // Compressed textures do not work in Kit 105.0 or 105.0.1
+    if (_kitVersion.rfind("105.0", 0) == 0) {
```
weegeekps, line 646: Is this even necessary? We just released version 0.11.0 which claims that versions less than 105.1 are unsupported.

weegeekps (commented):
One question but otherwise looks good.

src/core/src/Context.cpp
lilleyse: Oh good point... I'll remove it

lilleyse: I think I worked on this PR before realizing that 0.11.0 wouldn't be backwards compatible with Kit 105.0

weegeekps: In that case do we even need any of the code to pass the kit version up to the native layer?

lilleyse: Nope, I just removed it all

weegeekps (approved):
Looks great and works great.

PR #475, opened 2023-10-02
0.11.0 release

lilleyse:
Change log:

* **Breaking change:** Cesium for Omniverse now requires Kit 105.1 or above (USD Composer 2023.2.0 or above).
* Reduced the number of materials created when loading un-textured tilesets.
* Added debug option `cesium:debug:disableGeoreferencing` to `CesiumDataPrim` to disable georeferencing and view tilesets in ECEF coordinates.
* Improvements to C++ testing infrastructure.

Note: I temporarily removed nvrtc since https://github.com/CesiumGS/cesium-omniverse/pull/459 didn't make it into the release. I'll revert https://github.com/CesiumGS/cesium-omniverse/commit/ff727729d807b2d77e4b43169d1a5eb203d68d8d once the release is complete.

PR #459, opened 2023-08-28
Add foundational CUDA support and basic kernel

corybarr:
This PR add the foundational classes for adding CUDA kernels to tilesets. In this initial implementation, a selectable kernel is created per tile. Notes:

- If you want to see the placeholder kernel's output to verify it's working, uncomment the print statement in CudaKernels.h::createVoxelsKernel
- This PR runs only a placeholder kernel. The next PR will finalize handling the arguments non-trivial kernels need per type.
- There is a test USDA file that will start to load the Melbourne point-cloud dataset

[melbournePointCloudTileset.zip](https://github.com/CesiumGS/cesium-omniverse/files/12457480/melbournePointCloudTileset.zip)

PR #476, opened 2023-10-03
Revert "Temporarily remove nvrtc and nvrtc-builtins from install"

lilleyse:
Adds nvrtc and nvrtc-builtins back to the install. They were temporarily removed in https://github.com/CesiumGS/cesium-omniverse/pull/475 before the release since https://github.com/CesiumGS/cesium-omniverse/pull/459 hadn't been merged yet.


mattelser (approved):
LGTM

PR #479, opened 2023-10-03
Fix tbb library name on Linux

lilleyse:
This fixes some `tbb_debug.so.2 not found` errors when linking against debug kit-sdk.

Even though `tbb_debug.so` and `tbb_debug.so.2` exist in `omni.usd.libs` the extension only exposes `tbb_debug` for windows, so we need to link against plain `tbb` on Linux. Snippet from the `extension.toml` below:

```
[[native.library]]
"filter:platform"."windows-x86_64"."filter.config"."release"."path" = "bin/${lib_prefix}tbb${lib_ext}"
"filter:platform"."windows-x86_64"."filter:config"."debug"."path" = "bin/${lib_prefix}tbb_debug${lib_ext}"
```


weegeekps (approved):
LGTM

PR #480, opened 2023-10-03
Add additional search paths to all .kit apps

lilleyse:
This fixes the python debugging application, similar to the fix in https://github.com/CesiumGS/cesium-omniverse/pull/451.

I don't understand why `settings.app.exts` isn't being inherited from `cesium.omniverse.dev` but it doesn't hurt to copy-paste these lines into each app.


weegeekps (approved):
LGTM

PR #481, opened 2023-10-03
Reload tileset when disableTexturePool property changes

lilleyse:
Force the tileset to reload when the `disableTexturePool` property changes. Small omission from https://github.com/CesiumGS/cesium-omniverse/pull/389


weegeekps (approved):
LGTM

Issue #477, opened 2023-10-03
Application hangs when toggling "Disable" properties on CesiumDataPrim if CWT is loaded


corybarr:
To reproduce:

![image](https://github.com/CesiumGS/cesium-omniverse/assets/1040194/41f1e5f2-8f61-4df8-b36e-3d550674216c)

![image](https://github.com/CesiumGS/cesium-omniverse/assets/1040194/efbc3037-2495-480b-8b7e-0b67f8604a35)

lilleyse:
@corybarr do you know if this happens with earlier version of the extension, pre Kit 105.1?

lilleyse:
Still early to say, but adding `omnihydra.parallelHydraSprimSync = false` to `cesium.omniverse.dev.kit` almost completely fixes the hangs for me. And as a bonus, it doesn't show white tiles.

If the setting works for others I can open a PR to add it to `cesium.omniverse.dev.kit` and to the Python code.

corybarr:
I'm not able to hang the app with `omnihydra.parallelHydraSprimSync = false`. It's currently pretty easy for me to hang the app without it. For both of the hangs, the last console output was an error message about an Sprim:

`[omni.usd] Coding Error (secondary thread): in _MarkSprimDirty at line 699 of W:\ac88d7d902b57417\USD\pxr\imaging\hd\changeTracker.cpp -- Failed verification: ' it != _sprimState.end() '`

lilleyse:
I printed the carb settings for USD Composer 2023.2.0 and noticed that it's false there too. I feel pretty confident this is the right fix for us. I'll open a PR.

```
  "omnihydra": {
    "staticMaterialNetworkTopology": false,
    "parallelHydraSprimSync": false
  },
```

PR #482, opened 2023-10-04
Fix hangs when loading tileset materials

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/477

In Kit 105.1 we noticed that Omniverse would sometimes hang when loading tileset materials, i.e. the loading bar would get stuck at some %.

Setting `omnihydra.parallelHydraSprimSync` to `false` almost completely fixes these issues. A bonus is that materials no longer appear white/red when textures are loading.

I printed the carb settings for USD Composer 2023.2.0 and noticed that it's false there too, so I feel pretty confident this is the right fix for us. Since USD Composer already has the "fix" I don't think we need to do a patch release.

```
  "omnihydra": {
    "staticMaterialNetworkTopology": false,
    "parallelHydraSprimSync": false
  },
```


weegeekps (approved):
:ship:

PR #491, opened 2023-10-12
fix possible division by zero in object pool test

mattelser:
I'm surprised this wasn't tripped earlier, but this test can be (randomly) handed an empty pool, causing a division by zero.


weegeekps (approved):
LGTM

PR #492, opened 2023-10-16
Add support for multiple raster overlays

### LINKED ISSUE #84: Support for multiple raster overlays

lilleyse:
Fixes #84 

Adds support for multiple raster overlays.

* Works on untextured datasets like CWT and textured datasets like Google 3D Tiles
* Works for geographic and web mercator projections and future projections (i.e. texcoord sets are handled correctly)
* Works for raster overlays with alpha

This required changes in a few places:

* `FabricGeometry` and `FabricGeometryDefinition` now handle texcoord sets in a generic way. There were a few places where the texcoord set was hardcoded to zero. This has been fixed. See related changes in `GltfUtil`.
* `FabricMaterial` supports up to 10 texcoord sets and 16 textures per texture array. Excess texcoord sets and textures are ignored. When there's a single texture it constructs the same shader network as before, so no performance loss. When there's multiple textures it uses the new `cesium_texture_array_lookup` shader which is responsible for blending all the imagery layers together with a simple alpha over blend. See `cesium.mdl`.

Note that the changes to `FabricGeometryDefinition` and `FabricMaterialDefinition` affect pooling - i.e. geometries with the same number of texcoord sets will go to the same geometry pool and materials with the same number of base color textures will go the same material pool.

The main performance TODO is:

* `cesium_texture_array_lookup` is hardcoded for 16 textures which is much higher than the common case. If we procedurally generate the .mdl based on how many textures are in use it could improve performance. Or we could just have 16 versions of `cesium_texture_array_lookup` :shrug:.

Not supported yet:

* Styling, such as controlling the translucency of imagery layers

**Examples:**

[multiple.usda.zip](https://github.com/CesiumGS/cesium-omniverse/files/12922644/multiple.usda.zip)
Washington DC (Web Mercator) above Sentinel II (geographic) above Bing Maps (Web Mercator, not visible) 

image (11)

[multiple-google.usda.zip](https://github.com/CesiumGS/cesium-omniverse/files/12922646/multiple-google.usda.zip)
Washington DC (Web Meractor) above Google 3D Tiles (textured glTF)

image (12)

### DIFF
--- a/exts/cesium.omniverse/mdl/cesium.mdl
+++ b/exts/cesium.omniverse/mdl/cesium.mdl
@@ -11,6 +11,64 @@ module [[
     anno::display_name("Cesium MDL functions")
 ]];
 
-// For internal use only. See note in FabricMaterial.cpp
+// For internal use only
 export gltf_texture_lookup_value cesium_texture_lookup(*) [[ anno::hidden() ]] = gltf_texture_lookup();
 export material cesium_material(*) [[ anno::hidden() ]] = gltf_material();
+
+export gltf_texture_lookup_value cesium_texture_array_lookup(
+    uniform int texture_count = 0,
+    gltf_texture_lookup_value texture_0 = gltf_texture_lookup(),
+    gltf_texture_lookup_value texture_1 = gltf_texture_lookup(),
+    gltf_texture_lookup_value texture_2 = gltf_texture_lookup(),
+    gltf_texture_lookup_value texture_3 = gltf_texture_lookup(),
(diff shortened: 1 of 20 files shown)


weegeekps (approved):
Technically, this looks fine, but I question if this really solves the problem without controls regarding how much to blend and such. I'm going to approve but please consider that point. My gut is saying this doesn't actually solve the need.

PR #495, opened 2023-10-18
Multiple raster overlays cleanup

lilleyse:
As I was integrating multiple raster overlays ([#492](https://github.com/CesiumGS/cesium-omniverse/pull/492)) into tileset material nodes ([#406](https://github.com/CesiumGS/cesium-omniverse/pull/406)) I realized it would be much simpler to treat glTF base color textures and raster overlays as separate things instead of all being treated as base color textures.

The main change is in `cesium.mdl` where `cesium_imagery_layer_resolver` is responsible for blending imagery layers together and `cesium_material` is responsible for blending the final imagery layer color with the glTF base color texture. As a result `cesium_material` is no longer an exact copy of `gltf_material`, but does some simple calculations first.

In the case where there's a single imagery layer performance should still be good because it skips `cesium_imagery_layer_resolver`.

### DIFF
--- a/exts/cesium.omniverse/mdl/cesium.mdl
+++ b/exts/cesium.omniverse/mdl/cesium.mdl
@@ -11,64 +11,95 @@ module [[
     anno::display_name("Cesium MDL functions")
 ]];
 
-// For internal use only
-export gltf_texture_lookup_value cesium_texture_lookup(*) [[ anno::hidden() ]] = gltf_texture_lookup();
-export material cesium_material(*) [[ anno::hidden() ]] = gltf_material();
+float4 alpha_blend(float4 src, float4 dst) {
+    return src * float4(src.w, src.w, src.w, 1.0) + dst * (1.0 - src.w);
+}
 
-export gltf_texture_lookup_value cesium_texture_array_lookup(
-    uniform int texture_count = 0,
-    gltf_texture_lookup_value texture_0 = gltf_texture_lookup(),
-    gltf_texture_lookup_value texture_1 = gltf_texture_lookup(),
-    gltf_texture_lookup_value texture_2 = gltf_texture_lookup(),
-    gltf_texture_lookup_value texture_3 = gltf_texture_lookup(),
-    gltf_texture_lookup_value texture_4 = gltf_texture_lookup(),
-    gltf_texture_lookup_value texture_5 = gltf_texture_lookup(),
-    gltf_texture_lookup_value texture_6 = gltf_texture_lookup(),
-    gltf_texture_lookup_value texture_7 = gltf_texture_lookup(),
-    gltf_texture_lookup_value texture_8 = gltf_texture_lookup(),
-    gltf_texture_lookup_value texture_9 = gltf_texture_lookup(),
-    gltf_texture_lookup_value texture_10 = gltf_texture_lookup(),
-    gltf_texture_lookup_value texture_11 = gltf_texture_lookup(),
-    gltf_texture_lookup_value texture_12 = gltf_texture_lookup(),
-    gltf_texture_lookup_value texture_13 = gltf_texture_lookup(),
-    gltf_texture_lookup_value texture_14 = gltf_texture_lookup(),
-    gltf_texture_lookup_value texture_15 = gltf_texture_lookup()
-) [[ anno::hidden() ]]
-{
-    // The array length should match MAX_TEXTURE_LAYER_COUNT in Tokens.h
-    gltf_texture_lookup_value[16] texture_values(
-        texture_0,
-        texture_1,
-        texture_2,
-        texture_3,
-        texture_4,
-        texture_5,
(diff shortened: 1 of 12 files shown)


src/core/src/FabricPrepareRenderResources.cpp
```diff
@@ -56,22 +56,10 @@ struct TileLoadThreadResult {
 };
 
 bool hasBaseColorTextureGltf(const FabricMesh& fabricMesh) {
-    return fabricMesh.material->getMaterialDefinition().hasBaseColorTextures() &&
+    return fabricMesh.material != nullptr && fabricMesh.material->getMaterialDefinition().hasBaseColorTexture() &&
            fabricMesh.materialInfo.baseColorTexture.has_value();
```
lilleyse, line 60: Bug fix. The `nullptr` check prevent against a crash if you're using a custom tileset material but omniverse can't locate it (the full explanation for why this happens is a bit longer).

src/core/src/FabricMaterial.cpp
```diff
@@ -410,7 +407,7 @@ void FabricMaterial::setTextureValues(
     const TextureInfo& textureInfo,
     uint64_t texcoordIndex) {
 
-    if (texcoordIndex >= FabricTokens::MAX_TEXTURE_LAYER_COUNT) {
+    if (texcoordIndex >= FabricTokens::MAX_PRIMVAR_ST_COUNT) {
```
lilleyse, line 410: Bug fix, was using the wrong constant.

lilleyse (commented):
I pointed out some bug fixes that were included in this PR

src/core/src/FabricMaterial.cpp
```diff
@@ -320,67 +326,58 @@ void FabricMaterial::setMaterial(int64_t tilesetId, const MaterialInfo& material
     auto srw = UsdUtil::getFabricStageReaderWriter();
 
     setShaderValues(_shaderPath, materialInfo);
-    setTilesetId(tilesetId);
+
+    for (const auto& path : _allPaths) {
+        FabricUtil::setTilesetId(path, tilesetId);
```
corybarr, line 331: Would it ever be the case that a a prim and its path belongs to multiple tilesets? For example the shader in `_allPaths.push_back(shaderPath)`

exts/cesium.omniverse/mdl/cesium.mdl
```diff
@@ -11,64 +11,95 @@ module [[
...
+    gltf_texture_lookup_value imagery_layer_6 = gltf_texture_lookup(),
+    gltf_texture_lookup_value imagery_layer_7 = gltf_texture_lookup(),
+    gltf_texture_lookup_value imagery_layer_8 = gltf_texture_lookup(),
+    gltf_texture_lookup_value imagery_layer_9 = gltf_texture_lookup(),
+    gltf_texture_lookup_value imagery_layer_10 = gltf_texture_lookup(),
+    gltf_texture_lookup_value imagery_layer_11 = gltf_texture_lookup(),
+    gltf_texture_lookup_value imagery_layer_12 = gltf_texture_lookup(),
+    gltf_texture_lookup_value imagery_layer_13 = gltf_texture_lookup(),
+    gltf_texture_lookup_value imagery_layer_14 = gltf_texture_lookup(),
+    gltf_texture_lookup_value imagery_layer_15 = gltf_texture_lookup()
```
lilleyse, line 73: These should actually by `gltf_texture_lookup_value()`. It's harmless for now, but I fixed it on a different branch.

src/core/src/FabricMaterial.cpp
lilleyse: It shouldn't ever happen because unique IDs are given to every FabricGeometry / FabricMaterial that's created.

* https://github.com/CesiumGS/cesium-omniverse/blob/9137f544f67989daf3de0f8d587b4b71cbdd520a/src/core/src/FabricResourceManager.cpp#L74
* https://github.com/CesiumGS/cesium-omniverse/blob/9137f544f67989daf3de0f8d587b4b71cbdd520a/src/core/src/FabricGeometryPool.cpp#L26

lilleyse: And all the inner prims like shaders are appended to the material path.

corybarr: I think I see. "Creating" seems like what we'd call "cloning" in Unity.

exts/cesium.omniverse/mdl/cesium.mdl
lilleyse: Oh and I forgot to mention in the PR description, `eAsset` arrays don't seem to work in Fabric which is why each texture is its own entry instead of `gltf_texture_lookup_value[16]`

corybarr (approved):
I haven't spent much time with OV materials yet, but this all makes sense and LGTM.

Issue #429, opened 2023-08-16
`cesium:debug:randomColors` doesn't work if tile has vertex colors


lilleyse:
The `cesium:debug:randomColors` option added in https://github.com/CesiumGS/cesium-omniverse/pull/345 doesn't work if the tile has vertex colors.

There are a few different ways we could fix this:

* Option 1: Implement our own version of `gltf/pbr.mdl` that has separate uniforms for vertex color vs. tint
* Option 2: Multiply the debug color into the `base_color_factor`
* Option 3: Multiply the debug color into the vertex colors

I think option 2 would make the most sense in the short term.

PR #497, opened 2023-10-18
Fix `cesium:debug:randomColors` with vertex colors

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/429. Previously `randomColors` would have no affect for tiles with vertex colors (including point clouds).

Now that we have our own `cesium_material` shader I went with option 1 in https://github.com/CesiumGS/cesium-omniverse/issues/429.

Test data: [vertex-color.zip](https://github.com/CesiumGS/cesium-omniverse/files/13046516/vertex-color.zip)

Without `randomColors`|With `randomColors`
--|--
without|with

### DIFF
--- a/src/core/src/FabricGeometry.cpp
+++ b/src/core/src/FabricGeometry.cpp
@@ -153,7 +153,7 @@ void FabricGeometry::initialize() {
     }
 
     if (hasVertexColors) {
-        attributes.addAttribute(FabricTypes::primvars_vertexColor, FabricTokens::primvars_vertexColor);
+        attributes.addAttribute(FabricTypes::primvars_displayColor, FabricTokens::primvars_displayColor);
     }
 
     attributes.createAttributes(_path);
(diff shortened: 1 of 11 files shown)


src/core/src/FabricGeometry.cpp
```diff
@@ -158,7 +153,7 @@ void FabricGeometry::initialize() {
     }
 
     if (hasVertexColors) {
-        attributes.addAttribute(FabricTypes::primvars_vertexColor, FabricTokens::primvars_vertexColor);
```
corybarr, line 161: Is the vertexColor attribute never going to be used going forward for anything, such as styling?

corybarr: Ah, I see. Got it. Nevermind.

corybarr (approved):

src/core/src/FabricGeometry.cpp
lilleyse: Yeah it's basically just a rename from `vertexColor` to `displayColor`. Trying to be more consistent with USD conventions.

PR #499, opened 2023-10-20
Include conan.cmake to work around SSL issue in CI

lilleyse:
Thanks to @ptrgags for the original fix in a different repo:

> The last few CI builds have been failing due to an SSL cert problem when downloading `conan.cmake`. This PR follows [the advice here](https://github.com/conan-io/cmake-conan/issues/575#issuecomment-1770545072) to just download the file (it's small, only 36.5k) and include with the repo.


mattelser (approved):
LGTM

PR #500, opened 2023-10-20
Expand tests for UsdUtil, building out a bit more test infrastructure

mattelser:
Built out some more tests of `UsdUtil`. I feel like there's a way to remove repetition in prim/path creation/testing/removal, but I also suspect it would add more complication than it's worth at this stage. Also, happy to hear suggestions for better namespace organization.


exts/cesium.omniverse.cpp.tests/cesium/omniverse/cpp/tests/extension.py
```diff
@@ -3,12 +3,14 @@
 import omni.usd
 import omni.kit.ui
 import omni.kit.app
+from cesium.omniverse.utils import wait_n_frames
```
mattelser, line 6: whoops this can be removed

mattelser:
I forgot to clang-format a header and removed a redundant import. This should actually be ready for review now

exts/cesium.omniverse.cpp.tests/cesium/omniverse/cpp/tests/extension.py
```diff
@@ -30,12 +31,21 @@ def on_startup(self):
...
+            if not self.tests_set_up:
+                self.tests_set_up = True
+                print("Beginning Cesium Tests Extension tests")
+                stageId = omni.usd.get_context().get_stage_id()
+                tests_interface.set_up_tests(stageId)
+            else:
+                # unsubscribe so there's no way the next frame triggers another run
+                self._run_once_sub.unsubscribe()
+                tests_interface.run_all_tests()
+                print("Cesium Tests Extension tests complete")
```
weegeekps, line 48: Funky. Could we use a generator to solve this instead?

mattelser: we could but I think it would have to be used indirectly since we don't have control over how this function is called. We certainly could do something like:
```
myCoolTests:
    runSetup()
    yield foo
    runTests

myCoolTestsGlobal = myCoolTests()

run_once_after_stage_opens():
    if not next(myCoolTests, None):
         unsubscribeThisFunction()
```
But that also seems funky. I can't think of a non-funky way of doing this since we can't use `wait_n_frames`

weegeekps: Okay. I'm not going to block over this. I'll approve.

weegeekps (approved):
:ship:

exts/cesium.omniverse.cpp.tests/cesium/omniverse/cpp/tests/extension.py
mattelser: I've got to fix a (superfluous?) import to get the windows build working anyway, so if there's a cleaner way to accomplish this I'm happy to implement it

PR #502, opened 2023-10-20
Remove nvrtc from install step

lilleyse:
Removing the code that installs `nvrtc` to `exts/cesium.omniverse/bin` since we aren't using nvrtc yet. Instead I'm going to add this code to https://github.com/CesiumGS/cesium-omniverse/pull/459.


corybarr (approved):
Yeah, this strategy makes sense.

PR #503, opened 2023-10-20
Fix gdb pretty printing

lilleyse:
Since the switch from `lldb` to `cppdbg` in https://github.com/CesiumGS/cesium-omniverse/pull/478 I wasn't getting very useful information from the debugger (e.g. couldn't inspect std::vector elements, std::strings, etc)

I've had this problem before with `cppdbg` where the `-enable-pretty-printing` flag doesn't seem to have any effect.

I found a workaround that seems to do the trick.


corybarr (approved):
Linux-only so I can't actually test this, but it looks really straightforward.

PR #507, opened 2023-10-20
Automatically locate Cesium MDL exports in USD Composer

lilleyse:
This makes it possible for USD Composer to automatically locate `Cesium` related MDL functions.

image


corybarr (approved):
I haven't worked with MDL in OV, but this all makes sense and looks straighforward.

r-veenstra:
Confirming this worked on my fresh OS install with USD Composer 2023.2.0

PR #508, opened 2023-10-23
Update CHANGES.md

lilleyse:
We forgot to update `CHANGES.md` in a bunch of recent PRs.


corybarr (approved):
I'd be fine if it's very simple things to bypass a PR.

lilleyse:
@corybarr I sort of agree, but we'd need to disable branch protection rules to make that possible.

PR #509, opened 2023-10-23
Create onboarding and test documentation

mattelser:
Here is a bit of context and a few links I would have found useful to have in one place when I first started, as well as some documentation on the testing infrastructure.


lilleyse:
This looks great @mattelser. A documentation sweep is long overdue and this a good start.

lilleyse (approved):

Issue #510, opened 2023-10-23
Detect kit version and give useful error message if extension is incompatible


r-veenstra:
One of the most common issues from users is downloading versions of the extension that are incompatible with the version of kit being used by their Omniverse application.

Ideally there would be some way to detect the Kit version and prevent loading the extension if it is incompatible, along with a message explaining why.

While we maintain a USD composer compatibility list in our quickstart, the growing number of OV apps all running on different versions of kit will make that list hard to manage.

asluk:
Extensions can declare which versions of Kit they're compatible with - https://docs.omniverse.nvidia.com/kit/docs/kit-manual/latest/guide/extensions_advanced.html#package-target

r-veenstra:
Thanks @asluk, I took a quick look at this.  

We're currently using `kit = ["105.*"]`, however I can see a couple of our reported issues from users are still happening, likely due to a combination of factors within and not without our control.  See below.

- We have a breaking change between Kit 105.0 and Kit 105.1, so with our declared version of 105.* the current 105.1 only version of Cesium for Omniverse still tries to load in older Kit 105.0 apps.  @lilleyse should we just change this to be explicitely 105.1, and moving forward add future versions of Kit that have been tested?  Perhaps a touch more maintenance and testing, however that's not necessarily a bad thing.

- Most of the reported issues are coming from users trying to run Cesium for Omniverse in USD Composer 2022.3.3.  I believe this is because 2022.3.3 is still flagged as the **release** build in the Omniverse launcher, and shows up as the default version to install.  It's easy to overlook the availabiliy of newer **beta** builds (I made this mistake setting up OV on a fresh install of Windows, it will be less obvious for a new user).  This wouldn't be a huge issue, exept that 2022.3.3 doesn't actually respect the `kit = ["105.*"]` declaration, so it still tries to load much newer versions of the extension anyway.  I don't think there is much that can be actioned here, aside from waiting for a 2023.x version of USD Composer to be flagged the new release version, which will reduce the likelihood of users accidentally installing the older version in the first place.

lilleyse:
I played around with the `kit` target some more and noticed a few things:

* If the target kit version is greater, the extension doesn't show up in the list at all. I would expect an error message.
* If the target kit version is less, the extension is enabled. I would expect an error message.
* 2022.3.3 seems to respect the `kit = ["105.*"]` declaration, the error message is just not very descriptive

USD Composer Version|Kit version|Extension target|Expected|Result
[...]

PR #513, opened 2023-10-25
Force extension to require Kit 105.1

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/510, but with some caveats in https://github.com/CesiumGS/cesium-omniverse/issues/510#issuecomment-1777303188


corybarr (approved):
Very simple

PR #516, opened 2023-10-25
Fixing loading Globe Anchors. (again)

### LINKED ISSUE #512: Globe anchored prim disappears when changing latitude, longitude
When I change the latitude or longitude of a globe anchor the child prim dissappears. Interestingly, if I change the translation first and then change the latitude or longitude it works fine.

Tested with the sample data in https://github.com/CesiumGS/cesium-omniverse-samples/pull/8

https://github.com/CesiumGS/cesium-omniverse/assets/915398/ec4883f2-e208-4056-a180-e6ce4bf6b3ea

weegeekps:
Resolves #512.

The same issue that bit us in #494 wasn't fully fixed by that PR. Instead what should have been done, and is now being done, is that the call to recalculate the cache is made in the create call. This guarantees we have all the data we need on the next movement of the anchored prim.


lilleyse (approved):
Works nicely

PR #523, opened 2023-10-26
Replace regex attr name parsing with string parsing

corybarr:
When acquiring Fabric geometry, we were previously parsing attribute names with a regex. This costs a median of 244 microseconds. Replacing with string parsing drops the median to 107 microseconds. Stats via Superluminal for the FabricGeometryDefinition constructor:

Regex:

image

string:

image

### DIFF
--- a/src/core/src/GltfUtil.cpp
+++ b/src/core/src/GltfUtil.cpp
@@ -317,7 +317,7 @@ const CesiumGltf::ImageCesium& getImageCesium(const CesiumGltf::Model& model, co
 }
 
 std::pair<std::string, uint64_t> parseAttributeName(const std::string& attributeName) {
-    std::string semantic = "";
+    std::string semantic{};
     uint64_t setIndexU64 = 0;
 
     if (attributeName[0] != '_') {
--- a/src/core/src/GltfUtil.cpp
+++ b/src/core/src/GltfUtil.cpp
@@ -335,7 +335,10 @@ std::pair<std::string, uint64_t> parseAttributeName(const std::string& attribute
         semantic = attributeName;
     } else {
         semantic = attributeName.substr(0, lastUnderscorePosition);
-        std::from_chars(attributeName.data() + lastUnderscorePosition + 1, attributeName.data() + attributeName.size() - 1, setIndexU64);
+        std::from_chars(
+            attributeName.data() + lastUnderscorePosition + 1,
+            attributeName.data() + attributeName.size(),
+            setIndexU64);
     }
 
     return std::make_pair(semantic, setIndexU64);
--- a/src/core/src/GltfUtil.cpp
+++ b/src/core/src/GltfUtil.cpp
@@ -317,16 +317,23 @@ const CesiumGltf::ImageCesium& getImageCesium(const CesiumGltf::Model& model, co
 }
 
 std::pair<std::string, uint64_t> parseAttributeName(const std::string& attributeName) {
-    std::string semantic{};
-    uint64_t setIndexU64 = 0;
+    int searchPosition = static_cast<int>(attributeName.size()) - 1;
+    int lastUnderscorePosition{-1};
+    while (searchPosition > 0) {
+        if (!isdigit(attributeName[searchPosition])) {
+            break;
+        }
+        lastUnderscorePosition = searchPosition;
+        searchPosition--;
+    }
 
-    if (attributeName[0] != '_') {
+    std::string semantic{};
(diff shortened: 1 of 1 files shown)


src/core/src/GltfUtil.cpp
```diff
@@ -317,21 +317,19 @@ const CesiumGltf::ImageCesium& getImageCesium(const CesiumGltf::Model& model, co
...
-        setIndex = matches[3];
+    if (attributeName[0] != '_') {
+        semantic = attributeName;
+    } else {
+        auto lastUnderscorePosition = attributeName.rfind('_');
+        semantic = attributeName.substr(0, lastUnderscorePosition);
+        auto setIndex =
+            attributeName.substr(lastUnderscorePosition + 1, attributeName.size() - lastUnderscorePosition + 1);
+        std::from_chars(setIndex.data(), setIndex.data() + setIndex.size(), setIndexU64);
     }
```
lilleyse, line 331: I'm not sure if this check is robust enough. Some of the attributes below will work, but some will fail:

_FEATURE_ID_1
_CESIUMOVERLAY_1
TEXCOORD_1
POSITION
RANDOM_PROPERTY

A better solution might be to traverse backwards like, stopping either when:
* You hit an A-Z character (no set index)
* You hit an underscore (set index determined by preceding digits)

src/core/src/GltfUtil.cpp
```diff
@@ -317,21 +317,19 @@ const CesiumGltf::ImageCesium& getImageCesium(const CesiumGltf::Model& model, co
 }
 
 std::pair<std::string, uint64_t> parseAttributeName(const std::string& attributeName) {
-    const auto regex = std::regex("^([a-zA-Z_]*)(_([1-9]\\d*|0))?$");
-
-    std::string semantic;
-    std::string setIndex;
-
-    std::smatch matches;
+    std::string semantic = "";
```
weegeekps, line 320: I know it's not intuitive but you should just use `{}` instead. Much less expensive for an empty string. https://godbolt.org/z/nKx8c7EP9

```suggestion
    std::string semantic{};
```

src/core/src/GltfUtil.cpp
corybarr: Ah, good list. Is there a master list available? I assumed anything with a set index would start with '_', but it looks like TEXCOORD might not.

src/core/src/GltfUtil.cpp
corybarr: Ah yes, my recent multi-year stint in C# sometimes leaves me forgetting I can use direct initialization. Good call.

src/core/src/GltfUtil.cpp
lilleyse: This the official glTF list:

image

But custom semantics are allowed too (anything starting with an underscore)

corybarr: Good resource. Algorithm updated to meet the Khronos spec.

lilleyse:
Looks good. Did you re-run the profiler to see how the new implementation performs?

src/core/src/GltfUtil.cpp
```diff
@@ -317,20 +316,28 @@ const CesiumGltf::ImageCesium& getImageCesium(const CesiumGltf::Model& model, co
...
-    uint64_t setIndexU64 = 0;
-    std::from_chars(setIndex.data(), setIndex.data() + setIndex.size(), setIndexU64);
+    std::string semantic{};
+    uint64_t setIndexU64{0};
+    if (lastUnderscorePosition == -1) {
+        semantic = attributeName;
+    } else {
+        semantic = attributeName.substr(0, lastUnderscorePosition);
+        auto setIndex = attributeName.substr(lastUnderscorePosition + 1);
+        std::from_chars(setIndex.data(), setIndex.data() + setIndex.size(), setIndexU64);
```
lilleyse, line 339: It might be possible to avoid creating the `setIndex` string by passing in the offset / length to `std::from_chars`. One less heap allocation could help performance.

corybarr: Let's try that. It could only help.

corybarr: Without that heap allocation, it still levels out at 122 microseconds. It's probably a touch faster, but not enough to catch at the sampling rate. Still, let's keep it.

image

lilleyse (approved):

lilleyse:
@corybarr once you fix formatting this should be good to go

PR #527, opened 2023-10-30
added end-to-end tileset test

mattelser:
Created an end-to-end test for tileset loading, including the first c++ side event subscription in the codebase.


tests/src/tilesetTests.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <carb/dictionary/DictionaryUtils.h>
+#include <carb/events/IEvents.h>
+#include <doctest/doctest.h>
+#include <omni/kit/IApp.h>
+
+pxr::SdfPath endToEndTilesetPath;
+bool endToEndTilesetLoaded = false;
+carb::events::ISubscriptionPtr endToEndTilesetSubscriptionPtr;
+class TilesetLoadListener;
+TilesetLoadListener* tilesetLoadListener;
```
mattelser, line 18: It felt weird not to use a smart pointer, but for a global I didn't think they made sense

lilleyse: I would still switch to a `unique_ptr` just for internal consistency. In the cleanup function you can call `reset` on it to free the memory.

lilleyse (commented):
Looks good, just one comment

tests/src/tilesetTests.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+using namespace cesium::omniverse;
+
+class TilesetLoadListener final : public carb::events::IEventListener {
+  public:
+    size_t refCount = 0;
```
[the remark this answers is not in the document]
mattelser: It's mostly there so the virtual `addRef()` and `release()` have something to return

lilleyse (approved):

PR #528, opened 2023-10-30
Add power tool to save fabric stage

lilleyse:
Omniverse often crashes when saving the Fabric stage and then copying the contents of the textbox to somewhere else. A better workflow is to save the Fabric stage to a file.

I don't think printing the fabric stage needs to be part of the main extension, so I moved it to power tools.

https://github.com/CesiumGS/cesium-omniverse/assets/915398/3d8fc04e-1aed-4555-a109-a091d6a9b3a0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,5 +78,8 @@ tests/include/testUtils.h
 # Tracing files
 cesium-trace-*.json
 
-# Carb settings from Cesium Power Tools extensions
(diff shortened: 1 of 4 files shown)


weegeekps (approved):
LGTM

corybarr (approved):
Nice. I've hit this issue of printing the stage crashing OV.

PR #529, opened 2023-10-30
Add support for tileset displayColor and displayOpacity

lilleyse:
A tileset will now forward its `displayColor` and `displayOpacity` attributes over to fabric prims.

Also fixes a long running issue where vertex colors with alpha didn't work: https://github.com/CesiumGS/cesium-omniverse/pull/338

Note:
* You need to be in `RTX - Interactive` to see translucency. Otherwise any alpha less than 1.0 will appear completely black. There's a setting `fractionalCutoutOpacity` that's false by default for `RTX - Real Time`.
* Omniverse doesn't let you edit Display Color or Display Opacity through the UI, hence the Python script below
* These have no affect when using a tileset material. This seems to be consistent with general Omniverse behavior.

https://github.com/CesiumGS/cesium-omniverse/assets/915398/57429baf-ce2e-42af-a541-f795bda968cc

Script from the video:

```python
import omni.usd
import pxr.Gf

stage = omni.usd.get_context().get_stage()
tileset_prim = stage.GetPrimAtPath("/Cesium_World_Terrain")
tileset_prim.GetAttribute("primvars:displayColor").Set([pxr.Gf.Vec3f(0.0, 1.0, 0.0)])
tileset_prim.GetAttribute("primvars:displayOpacity").Set([0.5])
```

Datasets for testing:

* Vertex colors: [vertex-color.zip](https://github.com/CesiumGS/cesium-omniverse/files/13210593/vertex-color.zip)
* Regular tileset: [tileset-display-color-opacity.zip](https://github.com/CesiumGS/cesium-omniverse/files/13210597/tileset-display-color-opacity.zip)
* Point cloud rgba: [point-clouds-rgba.zip](https://github.com/CesiumGS/cesium-omniverse/files/13210644/point-clouds-rgba.zip) (voxel sizes are too big, unrelated to this PR)

image


exts/cesium.omniverse/mdl/cesium.mdl
```diff
@@ -203,18 +204,18 @@ float4 alpha_blend(float4 src, float4 dst) {
 
 float4 compute_base_color(
     gltf_texture_lookup_value imagery_layers_texture,
-    color debug_color,
+    float4 tile_color,
```
corybarr, line 207: Are we moving away from `fmt` for colors in general?

src/core/include/cesium/omniverse/Tokens.h
```diff
@@ -225,7 +227,7 @@ const omni::fabric::Type inputs_alpha_cutoff(omni::fabric::BaseDataType::eFloat,
 const omni::fabric::Type inputs_alpha_mode(omni::fabric::BaseDataType::eInt, 1, 0, omni::fabric::AttributeRole::eNone);
 const omni::fabric::Type inputs_base_alpha(omni::fabric::BaseDataType::eFloat, 1, 0, omni::fabric::AttributeRole::eNone);
 const omni::fabric::Type inputs_base_color_factor(omni::fabric::BaseDataType::eFloat, 3, 0, omni::fabric::AttributeRole::eColor);
-const omni::fabric::Type inputs_debug_color(omni::fabric::BaseDataType::eFloat, 3, 0, omni::fabric::AttributeRole::eColor);
+const omni::fabric::Type inputs_tile_color(omni::fabric::BaseDataType::eFloat, 4, 0, omni::fabric::AttributeRole::eNone);
```
corybarr, line 230: Why does the `AttributeRole` change?

exts/cesium.omniverse/mdl/cesium.mdl
corybarr: Ah, no alpha-channel support in `fmt`.

corybarr (approved):
Makes sense, looks good.

weegeekps:
> You need to be in RTX - Interactive to see translucency. Otherwise any alpha less than 1.0 will appear completely black. There's a setting fractionalCutoutOpacity that's false by default for RTX - Real Time.

Have we asked the Nvidia folks about if handling opacity in the real time renderer is on their radar?

src/core/include/cesium/omniverse/Tokens.h
lilleyse: My understanding (which could be wrong) is that `eColor` corresponds to the `color` MDL type, whereas `eNone` is a better fit for the `float4` MDL type. The `color` type doesn't seem to support alpha.

lilleyse: Also I haven't noticed any color space differences between `eColor` and `eNone`. Maybe linear is assumed for both, which works in our favor.

exts/cesium.omniverse/mdl/cesium.mdl
lilleyse: This is MDL code, so no relation to `fmt`

lilleyse:
> Have we asked the Nvidia folks about if handling opacity in the real time renderer is on their radar?

It seems to be disabled for performance reasons, but can be enabled dynamically by the app / extension.

Issue #525, opened 2023-10-27
Merge tests extension into main extension


mattelser:
Right now, as a direct result of being a separate extension, the tests extension executes in a separate memory space to the main extension. This doesn't present a problem for simpler interactions with the codebase, like directly calling functions to test them, however things might be get complicated when singletons get involved. @weegeekps and I looked into an issue today where we think this bit us. The working theory is that `AssetRegistry` (a singleton) was being called as expected in the two separate execution contexts, but that static variables were shared between the execution contexts as a result of the code calling the same dll/so. 

If we continue with a separate extension for tests, we would need hooks into the main extension for every operation that could run into singleton issues. I think a cleaner solution is to merge the two extensions, and enable all the test behavior via a build flag. This way, the main extension remains unaffected by tests when built for release, but the tests execute in the right context when enabled. 

I think this means:
- [ ] moving all the methods of `CesiumOmniverseCppTestsPlugin` into `CesiumOmniversePlugin`
- [ ] call `setUpTests()`, `runAllTests()`, and `cleanUpAfterTests()` from the main extension via a button, a flag, or something
- [ ] put the building of test functions behind a build flag
- [ ] put the calling of those functions on the python side behind that same build flag (or some other method of hiding them in release)
- [ ] update vs code build tasks to include the tests build flag
- [ ] update documentation

lilleyse:
Do the test extension and main extension need to be enabled at the same time? Can the test extension run on its own?

mattelser:
@lilleyse and I looked into this a bit. He found that the tests extension and main extension libraries are statically linked, which suggests that our theory about static variables in dynamically linked libraries probably doesn't explain the issue. 
Sean also created #530 to try removing the tests extension's dependency on the main extension. In testing that PR, I found there must be some amount of Cesium related initialization that has to happen beyond just getting `Context` since running the tests extension without the main extension failed the end-to-end tileset test. That initialization could possibly be on the python side, in the bindings, or some of the events broadcast by the main extension. We could certainly find and copy this initialization (or make it static and import/call it) to the tests extension, but the problem remains: at some point, in order to test everything in the codebase, we need to either merge parts of the main extension into the tests extension or merge the tests extension into the main extension.

PR #530, opened 2023-10-31
Remove cesium.omniverse dependency from tests extension

lilleyse:
May solve some of the issues brought up in https://github.com/CesiumGS/cesium-omniverse/issues/525


mattelser:
This works for all the tests currently on main, but I merged in the end-to-end tileset test in and that fails. Something else from the main extension must be required that the tests extension isn't setting up. I've added [a relevant comment to #525](https://github.com/CesiumGS/cesium-omniverse/issues/525#issuecomment-1787367725)

corybarr (approved):
straighforward

PR #532, opened 2023-10-31
Updated networking libraries and made some minor improvements

weegeekps:
There are several performance and security updates to `openssl`, `libcurl`, and `cpr` that we haven't kept up with. I've updated them all to the latest versions available on Conan for the tracks we're currently on. I did not update to the version 3 family of `openssl`, as that seems like a bigger bundle of changes that felt too significant for a small update that won't have much testing prior to public release.

That said, `openssl 1.1.1` is now at it's end-of-life, so we should consider upgrading to openssl 3 sometime in the coming months.


lilleyse (approved):
Looks good to me as long as Windows / Linux testing goes well.

weegeekps:
@lilleyse @corybarr Windows is fixed. It looks like the new version of either `libcurl` or `openssl` can't find the certificate store on Windows properly. We are now using the bundled certificates for both platforms.

corybarr:
> @lilleyse @corybarr Windows is fixed. It looks like the new version of either `libcurl` or `openssl` can't find the certificate store on Windows properly. We are now using the bundled certificates for both platforms.

@weegeekps Are you still thinking of holding off on getting it the upcoming release?

lilleyse:
I think we should hold off merging this until after the release. It would be good to do a profiling run on both windows and linux.

weegeekps:
I agree with @lilleyse. Additionally I think we should probably do a bit more general use testing or it to see if we run into any other issues.

lilleyse (approved):

PR #534, opened 2023-11-01
0.13.0 release

lilleyse:
* Changing certain tileset properties no longer triggers a tileset reload.
* Added support for `displayColor` and `displayOpacity` for tileset prims.
* Fixed rendering point clouds with alpha values.


weegeekps (approved):
LGTM

PR #535, opened 2023-11-01
Update to cesium-native v0.29.0

lilleyse:
Update cesium-native so that we get access to the new metadata API.

See https://github.com/CesiumGS/cesium-native/blob/main/CHANGES.md for the full list of changes.


corybarr (approved):
Straightforward

PR #536, opened 2023-11-01
Remove doubleSided from FabricGeometryDefinition

lilleyse:
This is a cleanup thing I've been meaning to fix for a while.

`doubleSided` doesn't really belong in `FabricGeometryDefinition`. The purpose of `FabricGeometryDefinition` is to prevent geometry with different attributes from sharing the same pool. Important for things like vertex colors, normals, and texcoords which affect the prim's topology, but not important for `doubleSided` which always exists on the prim.


corybarr (approved):
Yeah, that makes sense to move `doubleSided` out.

PR #537, opened 2023-11-02
Migrate to AlmaLinux 8 for builds

### LINKED ISSUE #464: Switch to Alma Linux 8 for release builds / Docker
CentOS 7 is approaching End of Life and we should switch to a different OS for our release builds. The minimum recommendations we received were

* Ubuntu: Minimum - 20.04
* RHEL: Minimum - 8.8
* Rocky: Minimum - 8

Ubuntu 20.04 uses glibc 2.31 and RHEL 8.8 and Rocky 8 use glibc 2.28. So with should go with one of the latter two for the widest glibc compatibility.

@weegeekps suggested Alma Linux 8 as an alternative to RHEL and Rocky.

We should update the Docker image and GitHub Actions before the October release.

weegeekps:
Resolves #464.

Moves CI builds to using our new AlmaLinux 8 based images.

There's been a few changes to how our OS Dockerfile is configured. Most of these are related to moving from CentOS 7 to AlmaLinux 8, such as using `dnf`, and how to select the 535 branch of the Nvidia drivers. However one important one is this:

```
# Enables gcc 11 for use within the docker image.
RUN echo "source /opt/rh/gcc-toolset-11/enable" >> /etc/bashrc
SHELL ["/bin/bash", "--login", "-c"]
```

Within the RHEL ecosystem `update-alternatives` doesn't exist (it was added in RHEL 7 as a helpful alias to `alternatives`, which is similar but not the same as the Ubuntu alternatives system), and it's generally considered not great to use within RHEL. The right way to enable the `gcc-toolset` is to source `enable`. Docker makes this a bit more difficult but you can just set `SHELL` to be a login shell and it should all work okay.

Python 3.9 is already the default in AlmaLinux 8, so we don't need to install it anymore as well.

### DIFF
--- a/.github/actions/action.yml
+++ b/.github/actions/action.yml
@@ -16,7 +16,8 @@ runs:
     # Parallel 2 because github action limited to two cores
     # Linters are turned off because there is no pre-compiled C++20-capable LLVM toolset for CentOS 7. It won't affect the build, it just means there won't be code formatting or linting. It will build fine with GCC.
     # Make sure to configure without sanitizers and any other development options
-    - "cmake -B build -D CMAKE_BUILD_TYPE=${{ inputs.build-type }} -D CESIUM_OMNI_ENABLE_LINTERS=OFF -D CESIUM_OMNI_ENABLE_SANITIZERS=OFF &&
+    - "git config --global --add safe.directory '*' &&
+      cmake -B build -D CMAKE_BUILD_TYPE=${{ inputs.build-type }} -D CESIUM_OMNI_ENABLE_LINTERS=OFF -D CESIUM_OMNI_ENABLE_SANITIZERS=OFF &&
       cmake --build build --parallel 2 &&
       cmake --build build --target install &&
       cmake --build build --target package"
--- a/.github/actions/action.yml
+++ b/.github/actions/action.yml
@@ -1,5 +1,5 @@
-name: "CentOS 7 Build"
-description: "Build Cesium for Omniverse in Docker container running a CentOS 7 image"
+name: "AlmaLinux 8 Build"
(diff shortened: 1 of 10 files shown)


weegeekps:
@lilleyse I think you'll need to do some stuff to disable CentOS 7 so we can merge this. Also I set it up to squash & merge to squash everything into a single commit that will be easier for us to back out of.

Removing the CentOS 7 docker image will come in a separate PR that I'll open once this is merged in. I am also going to look into what it will take to decommission the repo we have for the centos 7 images.

lilleyse:
> Removing the CentOS 7 docker image will come in a separate PR that I'll open once this is merged in

While you're at it could you do a documentation sweep? There are still many mentions of CentOS throughout the repo.

lilleyse:
> @lilleyse I think you'll need to do some stuff to disable CentOS 7 so we can merge this. Also I set it up to squash & merge to squash everything into a single commit that will be easier for us to back out of.

Done

lilleyse:
Have you verified that the published builds work in USD Composer?

lilleyse:
Do you know if CMake 3.24 or above is available, or do we still need to compile it ourselves?

weegeekps:
> While you're at it could you do a documentation sweep? There are still many mentions of CentOS throughout the repo.

Good call, working on that now.

> Have you verified that the published builds work in USD Composer?

Yup. Works great. Sorry, for not adding that to my comment a bit ago.

> Do you know if CMake 3.24 or above is available, or do we still need to compile it ourselves?

That was one of the first things I checked yesterday. Unfortunately, the latest version of CMake available in the AlmaLinux 8 repositories is 3.20.

lilleyse (approved):

PR #538, opened 2023-11-03
Add support for custom vertex attributes

lilleyse:
Added support for custom vertex attributes. Custom attributes can be accessed by their glTF attribute name with the primvar lookup functions in MDL, e.g. `data_lookup_int` and `data_lookup_float`.

In this example, the MDL is styling the point cloud based on the `_CLASSIFICATION` and `_INTENSITY` attributes from the glTF. The data is from [here](https://github.com/CesiumGS/3d-tiles-samples/tree/main/glTF/EXT_structural_metadata/PropertyAttributesPointCloud). The blockiness is due to the default voxel size being too large for this dataset.

image

[property-attributes.zip](https://github.com/CesiumGS/cesium-omniverse/files/13253251/property-attributes.zip)

Unfortunately this approach doesn't work for most pnts datasets because cesium-native transcodes pnts to `EXT_mesh_features` / property tables instead of property attributes, so per-point properties aren't accessible as vertex attributes. But I plan on adding support for property tables in a future PR.

While there's a lot of code changes the approach is pretty simple: add custom attributes to `FabricGeometryDefinition` and then add the values to `FabricGeometry`. This involved a little bit of template magic.

For consistency, vertex colors are now accessed with the name `COLOR_0` instead of `vertexColor`.

### DIFF
--- a/src/core/include/cesium/omniverse/GltfAccessors.h
+++ b/src/core/include/cesium/omniverse/GltfAccessors.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#include "cesium/omniverse/VertexAttributeType.h"
+
 #ifdef CESIUM_OMNI_MSVC
 #pragma push_macro("OPAQUE")
 #undef OPAQUE
@@ -89,7 +91,7 @@ class VertexColorsAccessor {
     VertexColorsAccessor(const CesiumGltf::AccessorView<glm::fvec3>& float32Vec3View);
     VertexColorsAccessor(const CesiumGltf::AccessorView<glm::fvec4>& float32Vec4View);
 
-    void fill(const gsl::span<glm::fvec4>& values) const;
+    void fill(const gsl::span<glm::fvec4>& values, uint64_t repeat = 1) const;
     [[nodiscard]] uint64_t size() const;
 
   private:
@@ -112,6 +114,20 @@ class FaceVertexCountsAccessor {
     [[nodiscard]] uint64_t size() const;
 
   private:
+    uint64_t _size;
+};
+
+template <VertexAttributeType T> class VertexAttributeAccessor {
+  public:
+    VertexAttributeAccessor();
+    VertexAttributeAccessor(const CesiumGltf::AccessorView<GetNativeType<T>>& view);
+
+    void fill(const gsl::span<GetFabricType<T>>& values, uint64_t repeat = 1) const;
+
+    [[nodiscard]] uint64_t size() const;
+
+  private:
+    CesiumGltf::AccessorView<GetNativeType<T>> _view;
     uint64_t _size;
 };
 
--- a/src/core/include/cesium/omniverse/GltfAccessors.h
+++ b/src/core/include/cesium/omniverse/GltfAccessors.h
@@ -91,7 +91,14 @@ class VertexColorsAccessor {
     VertexColorsAccessor(const CesiumGltf::AccessorView<glm::fvec3>& float32Vec3View);
     VertexColorsAccessor(const CesiumGltf::AccessorView<glm::fvec4>& float32Vec4View);
 
+    /**
+     * @brief Copy accessor values to the given output values, including any data transformations.
+     *
+     * @param values The output values.
+     * @param repeat Indicates how many times each value in the accessor should be repeated in the output. Typically repeat is 1, but for voxel point clouds repeat is 8.
+     */
     void fill(const gsl::span<glm::fvec4>& values, uint64_t repeat = 1) const;
+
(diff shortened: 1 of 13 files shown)


src/core/include/cesium/omniverse/GltfAccessors.h
```diff
@@ -89,7 +91,7 @@ class VertexColorsAccessor {
     VertexColorsAccessor(const CesiumGltf::AccessorView<glm::fvec3>& float32Vec3View);
     VertexColorsAccessor(const CesiumGltf::AccessorView<glm::fvec4>& float32Vec4View);
 
-    void fill(const gsl::span<glm::fvec4>& values) const;
+    void fill(const gsl::span<glm::fvec4>& values, uint64_t repeat = 1) const;
```
weegeekps, line 94: NIT: We should probably add documentation here while we're at it.

src/core/src/FabricGeometryDefinition.cpp
```diff
@@ -46,6 +51,10 @@ bool FabricGeometryDefinition::operator==(const FabricGeometryDefinition& other)
         return false;
     }
 
+    if (_customVertexAttributes != other._customVertexAttributes) {
+        return false;
```
weegeekps, line 55: We really don't need the branching in this operator? Why don't we just do...

```
    return _hasNormals == other._hasNormals && 
        _hasVertexColors == other._hasVertexColors && 
        _textcoordSetCount == other._texcoordSetCount &&
        _customVertexAttributes == other._customVertexAttributes
```

...instead? This suggestion seems clearer to me because it removes ambiguity that we want the all of these values to be equal.

src/core/src/VertexAttributeType.cpp
```diff
@@ -0,0 +1,113 @@
...
+        case VertexAttributeType::VEC3_FLOAT32:
+            return 3;
+        case VertexAttributeType::VEC4_UINT8:
+        case VertexAttributeType::VEC4_INT8:
+        case VertexAttributeType::VEC4_UINT16:
+        case VertexAttributeType::VEC4_INT16:
+        case VertexAttributeType::VEC4_FLOAT32:
+            return 4;
+        default:
+            assert(false);
```
weegeekps, line 36: If this is an error case we should probably also log an error to the console so we can diagnose it if someone runs into it.

weegeekps (commented):
A few comments.

lilleyse:
> Also, I did run this locally and it ran fine, but didn't test out the functionality as I don't have good test data handy.

That reminds me, I forgot to upload the .usda I was testing with. I added it to the PR description.

src/core/src/FabricGeometryDefinition.cpp
lilleyse: Done. Unfortunately clang-format doesn't place them nicely on separate lines.

src/core/include/cesium/omniverse/GltfAccessors.h
lilleyse: Done, but only for this function. We'll need to do a doc pass at some other point.

src/core/src/VertexAttributeType.cpp
lilleyse: I'm not sure if it's worth logging an error here since this code can only be reached if `type` not one of the enum values. The only time this could happen is if we reinterpet_cast an arbitrary int to `VertexAttributeType`. It's extremely unlikely that would happen, even unintentionally.

I did move the assert outside the switch/case though because we don't want anything falling through to the `default `case.

lilleyse:
@weegeekps with you were reviewing I realized that I wasn't handling normalized attributes. I pushed that change in https://github.com/CesiumGS/cesium-omniverse/pull/538/commits/b4d42ad3f65db4d6a919b8964bcaee7028a6efb6.

lilleyse:
After a small change, feature ids can now be accessed in MDL as `_FEATURE_ID_0` etc

feature-id

There's one caveat though: even though all feature ids in a triangle are the same value, the interpolation math can yield non integer results (e.g. 0.99999999 or 1.00000001), which when floored result in different feature ids across the surface. Here's what that looks like:

image

The fix is to read the primvar as a float and round the value. I included both versions in the zip file.

[feature-id-attribute.zip](https://github.com/CesiumGS/cesium-omniverse/files/13254098/feature-id-attribute.zip) - sample data from [3d-tiles-samples](https://github.com/CesiumGS/3d-tiles-samples/tree/main/glTF/EXT_mesh_features/FeatureIdAttribute)

weegeekps (approved):
Works great.

Issue #504, opened 2023-10-20
Update tiles automatically when material changes


lilleyse:
Tileset materials were added in https://github.com/CesiumGS/cesium-omniverse/pull/406 but there are some limitations. When you edit a tileset material the changes aren't propagated back to the tiles. There are two types of edits that we need to handle:

* [x] Changing individual material values - this could be implemented by watching for USD notifications and updating Fabric prims (a bit like https://github.com/CesiumGS/cesium-omniverse/pull/498).
* [ ] Changes to the node graph - this is more involved and probably requires USD notification + tileset reload due to the fact that every tile's material is a copy of the material being edited.

lilleyse:
Alternatively, we could detect when a material is changed and prompt the user to reload the tileset rather than doing it automatically, kind of like the "Fetch USD changes" popup that appears when you edit a USD file. That might be less frustrating.

<img width="598" alt="image" src="https://github.com/CesiumGS/cesium-omniverse/assets/915398/a9e1321b-c6a4-4695-84a1-26423549fd4d">

PR #540, opened 2023-11-09
Clean up FabricMaterial

lilleyse:
I did some cleanup in `FabricMaterial` while adding support for feature ids.

* The base color texture is now set in `setMaterial` rather than a separate function. This keeps the material setting code more cohesive. Imagery layers are still set separately because their loading happens in a different phase of the pipeline.
* Reverted some of the changes in https://github.com/CesiumGS/cesium-omniverse/pull/406. E.g. if the user has multiple `cesium_base_color_texture_float4` nodes, we no longer create multiple base color texture prims to feed into each one. Instead we create a single prim and add multiple connections. This is cleaner and should be more performant. This applies to imagery layers too.
* General cleanup that made it easier to implement feature ids, and might make it easier to  fix https://github.com/CesiumGS/cesium-omniverse/issues/504

I tested this PR with a variety of scenes from previous PRs: [test-data-sets.zip](https://github.com/CesiumGS/cesium-omniverse/files/13302959/test-data-sets.zip)

### DIFF
--- a/src/core/include/cesium/omniverse/FabricMaterial.h
+++ b/src/core/include/cesium/omniverse/FabricMaterial.h
@@ -14,6 +14,8 @@ class DynamicTextureProvider;
 
 namespace cesium::omniverse {
 
+class FabricTexture;
+
 class FabricMaterial {
   public:
     FabricMaterial(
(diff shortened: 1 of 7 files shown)


src/core/src/FabricPrepareRenderResources.cpp
```diff
@@ -55,9 +55,8 @@ struct TileLoadThreadResult {
     glm::dmat4 tileTransform;
 };
 
-bool hasBaseColorTextureGltf(const FabricMesh& fabricMesh) {
-    return fabricMesh.material != nullptr && fabricMesh.material->getMaterialDefinition().hasBaseColorTexture() &&
-           fabricMesh.materialInfo.baseColorTexture.has_value();
+bool hasBaseColorTexture(const FabricMesh& fabricMesh) {
+    return fabricMesh.material != nullptr && fabricMesh.material->getMaterialDefinition().hasBaseColorTexture();
 }
```
lilleyse, line 60: The deleted part used to be important, but is no longer needed after https://github.com/CesiumGS/cesium-omniverse/pull/495.

src/core/include/cesium/omniverse/FabricMaterial.h
```diff
@@ -3,6 +3,11 @@
 #include "cesium/omniverse/FabricMaterialDefinition.h"
 #include "cesium/omniverse/GltfUtil.h"
 
+#ifdef CESIUM_OMNI_MSVC
+#pragma push_macro("OPAQUE")
+#undef OPAQUE
```
corybarr, line 8: Is OPAQUE ever defined?

corybarr (approved):

PR #541, opened 2023-11-09
Add support for EXT_mesh_features

lilleyse:
Adds support for [EXT_mesh_features](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_mesh_features) and adds an MDL function `cesium_feature_id_int` that hides all the complex implementation details.

Three types of feature ID modes are supported:

* **Per-index feature IDs**: this is implemented with a `vertexId` primvar that mimicks OpenGL's `gl_VertexId`. This will  eventually be important for doing point cloud styling.
* **Per-vertex feature IDs**: similar to above, but reads the `_FEATURE_ID_n` primvar (made possible by [#538](https://github.com/CesiumGS/cesium-omniverse/pull/538)).
* **Per-texel feature IDs**: load feature ID textures. Made a few changes to `FabricTexture` to support linear vs. sRGB transfer functions and added a `texel_fetch` helper function in `cesium.mdl` to do nearest sampling (which unfortunately also forces us to implement our own texture wrapping code).

The finer points of the extension are also supported, like handling arbitrary channels and `nullFeatureId`.

feature-id-texture-ferry-building

[ferry-building.usda.zip](https://github.com/CesiumGS/cesium-omniverse/files/13302930/ferry-building.usda.zip)

What's cool about this test scene is that it uses almost all the cesium.mdl functions that we export. It takes a really long time to compile though.

### DIFF
--- a/exts/cesium.omniverse/mdl/cesium.mdl
+++ b/exts/cesium.omniverse/mdl/cesium.mdl
@@ -4,6 +4,7 @@ import ::anno::*;
 import ::state::*;
 import ::tex::*;
 import ::scene::*;
+import ::math::*;
 
 using ::gltf::pbr import *;
 
@@ -13,6 +14,8 @@ module [[
 ]];
 
 annotation annotation_not_connectable();
+
+export const auto DEFAULT_NULL_FEATURE_ID = -1;
 
 export enum up_axis_mode {
     Y,
@@ -124,6 +127,27 @@ export float4 cesium_imagery_layer_float4(
     return imagery_layer.valid ? imagery_layer.value : float4(0.0);
 }
 
+export int cesium_feature_id_int(
+    int feature_id = DEFAULT_NULL_FEATURE_ID
+    [[
+        anno::hidden(),
+        annotation_not_connectable()
+    ]],
+    uniform int feature_id_set_index = 0
+    [[
+        anno::unused()
+    ]]
+)
+[[
+    anno::display_name("Cesium feature ID lookup int"),
+    anno::description("Returns the feature ID for the given feature ID set. Returns -1 if the feature ID set does not exist or the feature ID is equal to nullFeatureId in the EXT_mesh_features extension."),
+    anno::author("Cesium GS Inc."),
+    anno::in_group("Cesium")
+]]
+{
+    return feature_id;
+}
+
 export material cesium_material(
     color base_color_factor = color(1.0)
     [[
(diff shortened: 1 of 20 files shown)


src/core/src/FabricPrepareRenderResources.cpp
```diff
@@ -149,6 +182,51 @@ std::vector<FabricMesh> acquireFabricMeshes(
...
+            uint64_t seenIndexes = 0;
+            for (uint64_t setIndex = 0; setIndex < featuresInfo.featureIds.size(); setIndex++) {
+                if (std::holds_alternative<uint64_t>(featuresInfo.featureIds[setIndex].featureIdStorage)) {
+                    if (seenIndexes++ == i) {
+                        fabricMesh.featureIdIndexIndexMapping.push_back(setIndex);
+                    }
+                }
+            }
+        }
+        assert(fabricMesh.featureIdIndexIndexMapping.size() == featureIdIndexCount);
```
lilleyse, line 229: I plan on cleaning this up soon. It can be more concise.

exts/cesium.omniverse/mdl/cesium.mdl
```diff
@@ -220,6 +244,173 @@ float4 compute_base_color(
     return base_color;
 }
 
+// Copied from gltf/pbr.mdl since it's not exported
```
corybarr, line 247: I've done very similar things. What happens if `gltf/pbr.mdl` is updated? (I never found a good solution to this question.)

src/core/src/FabricMaterial.cpp
```diff
@@ -26,6 +26,51 @@ const auto DEFAULT_DISPLAY_OPACITY = 1.0;
...
+    uint64_t attributeCount;
+    uint64_t textureCount;
+    uint64_t totalCount;
+};
+
+FeatureIdCounts getFeatureIdCounts(const FabricMaterialDefinition& materialDefinition) {
+    const auto& featureIdTypes = materialDefinition.getFeatureIdTypes();
+    auto featureIdCount = featureIdTypes.size();
+
+    if (featureIdCount > FabricTokens::MAX_FEATURE_ID_COUNT) {
```
corybarr, line 43: Is 10 an arbitrary limit?

src/core/include/cesium/omniverse/FabricMaterial.h
```diff
@@ -95,6 +103,15 @@ class FabricMaterial {
         uint64_t texcoordIndex,
         double alpha);
     void setImageryLayerAlphaValue(const omni::fabric::Path& path, double alpha);
+    void setFeatureIdIndexValues(const omni::fabric::Path& path, int nullFeatureId);
+    void
```
corybarr, line 107: This is very nit, but this seems like an odd way to divide up a line. Is this clang? The other function declaration and split at the parameter declarations.

lilleyse: Yes it's clang-format :slightly_frowning_face:

src/core/src/FabricPrepareRenderResources.cpp
```diff
@@ -168,7 +191,15 @@ void setFabricTextures(
...
-            baseColorTexture->setImage(*baseColorTextureImage);
+            baseColorTexture->setImage(*baseColorTextureImage, TransferFunction::SRGB);
+        }
+
+        const auto featureIdTextureCount = getFeatureIdTextureCount(mesh);
+        for (uint64_t j = 0; j < featureIdTextureCount; j++) {
+            const auto featureIdSetIndex = mesh.featureIdTextureSetIndexMapping[j];
+            const auto featureIdTextureImage = GltfUtil::getFeatureIdTextureImage(model, primitive, featureIdSetIndex);
+            assert(featureIdTextureImage);
+            mesh.featureIdTextures[j]->setImage(*featureIdTextureImage, TransferFunction::LINEAR);
```
corybarr, line 202: So the feature ID textures are compressed differently at tileset creation time because they're data not meant for direct visualization

exts/cesium.omniverse/mdl/cesium.mdl
lilleyse: I don't have a good solution either. We're absorbing so much of `gltf/pbr.mdl` at this point that I wouldn't be surprised if we copy the whole file.

src/core/src/FabricMaterial.cpp
lilleyse: Yeah kind of, let me see if the limit can be removed...

src/core/src/FabricPrepareRenderResources.cpp
lilleyse: Yeah, the PNGs themselves are encoded with a linear transfer function (as the spec requires)

corybarr (approved):

src/core/src/FabricMaterial.cpp
lilleyse: I may do that in a follow up PR

PR #543, opened 2023-11-13
Enabling caching for packman

### LINKED ISSUE #522: Improve CI speed
CI takes about 20 minutes to run, it be nice to cut this down back to 10-15 minutes.

The slowest part seems to be downloading NVIDIA libraries, which are listed here:

* https://github.com/CesiumGS/cesium-omniverse/blob/main/extern/nvidia/deps/kit-sdk.packman.xml
* https://github.com/CesiumGS/cesium-omniverse/blob/main/extern/nvidia/deps/target-deps.packman.xml

We could do some things to improve that like not downloading kit-sdk-debug in CI builds. Or we could cache the download with GitHub Actions (assuming we can download it faster from GitHub's servers than Nvidia's servers).

weegeekps:
Resolves #522.

Enables caching for Nvidia packman. Caching is performed by platform, not job, so Ubuntu and AlmaLinux share caches. This keeps us under GitHub's 10 GB limit on cumulative cache size.

The hash is generated on both `*.packman.xml` files since all of the artifacts go into the same folder.

Kit Debug builds are no longer downloaded on the build servers. I am abusing Packman user files to do this, but Packman has no other mechanisms for handling this sort of build flag optional requirements.

### DIFF
--- a/.github/actions/action.yml
+++ b/.github/actions/action.yml
@@ -10,7 +10,7 @@ runs:
   # Tell Conan to look for or create its build folder (.conan) in the repository's root directory
   env:
     CONAN_USER_HOME: $GITHUB_WORKSPACE
-    PACKMAN_ROOT: $GITHUB_WORKSPACE/packman-repo
+    PM_PACKAGES_ROOT: $GITHUB_WORKSPACE/packman-repo
   args:
     # Tell bash to run a command
     - "-c"
--- a/.github/actions/action.yml
+++ b/.github/actions/action.yml
@@ -10,6 +10,7 @@ runs:
   # Tell Conan to look for or create its build folder (.conan) in the repository's root directory
   env:
     CONAN_USER_HOME: $GITHUB_WORKSPACE
+    PACKMAN_ROOT: $GITHUB_WORKSPACE/packman-repo
   args:
     # Tell bash to run a command
     - "-c"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -86,7 +86,7 @@ jobs:
         id: packman-cache
         uses: actions/cache@v3
         with:
-          path: ${{ github.workspace }}/extern/nvidia/_build
+          path: extern/nvidia/_build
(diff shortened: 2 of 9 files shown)


weegeekps:
Caching Packman works great! This is a problem though.

image

.github/actions/action.yml
```diff
@@ -10,6 +10,7 @@ runs:
   # Tell Conan to look for or create its build folder (.conan) in the repository's root directory
   env:
     CONAN_USER_HOME: $GITHUB_WORKSPACE
+    PACKMAN_ROOT: $GITHUB_WORKSPACE/packman-repo
```
lilleyse, line 13: Is this env variable used?

weegeekps: Oops, that's an error. Let me fix quickly.

lilleyse:
I'm getting an error when running this locally with `-DCESIUM_OMNI_USE_NVIDIA_DEBUG_LIBRARIES`

```
Performing file copy with glob pattern "*.user.xml"
        Source: /home/slilley/Code/cesium-omniverse-2/extern/nvidia/debug-deps
        Destination: /home/slilley/Code/cesium-omniverse-2/extern/nvidia/deps

Traceback (most recent call last):
  File "/home/slilley/Code/cesium-omniverse-2/scripts/copy_from_dir.py", line 40, in <module>
    sys.exit(main())
  File "/home/slilley/Code/cesium-omniverse-2/scripts/copy_from_dir.py", line 27, in main
    source_files = glob.glob(glob_pattern, root_dir=source_dir)
TypeError: glob() got an unexpected keyword argument 'root_dir'
```

weegeekps:
> I'm getting an error when running this locally with `-DCESIUM_OMNI_USE_NVIDIA_DEBUG_LIBRARIES`

Which platform?

lilleyse:
Linux

weegeekps:
Okay, this was caused by the `root_dir` argument in `glob.glob` which requires Python 3.10. I switched to using the glob function in `pathlib` instead.

lilleyse (approved):
# :rocket:

PR #544, opened 2023-11-13
Fix startup crash due to global variable initialization order

lilleyse:
Fixes a bug where global variables in `FabricMaterial.cpp` were being initialized before dependent global variables in `GltfUtil.cpp`. 

Since the order of global variable initialization across translation units is undefined, the quick fix was to remove those global variables from `FabricMaterial.cpp`.

This bug was recently introduced by https://github.com/CesiumGS/cesium-omniverse/pull/540, so no `CHANGES.md` update required here.


corybarr (approved):
Makes sense

PR #545, opened 2023-11-13
Cleanup prim creation in FabricMaterial.cpp

lilleyse:
Created a helper function `createAttributes` that moves a lot of the common attribute initialization code into a single place.


weegeekps (approved):
Nice change.

PR #547, opened 2023-11-14
Pinning mypy to version 1.6.1

weegeekps:
Mypy just released a new version of the stub generator which has some [major regressions since mypy 1.6.1](https://github.com/python/mypy/issues/16486). In order to fix this, and prevent this from happening again, we should pin ourselves to version 1.6.1 which we know works. I tested this out again and verified that our genStubs script works correctly with the pinned version.


corybarr (approved):
Nice and simple and clean

PR #548, opened 2023-11-15
Add `package.writeTarget`

lilleyse:
Added `package.writeTarget` to `extension.toml` so that the extension gets automatically published.

https://docs.omniverse.nvidia.com/kit/docs/kit-manual/latest/guide/extensions_advanced.html#package-writetarget


weegeekps (approved):
approved-stamp-of-approval

PR #552, opened 2023-11-21
Remove harcoded limit for feature id sets and texcoord sets

lilleyse:
Merge https://github.com/CesiumGS/cesium-omniverse/pull/545 first

Addresses https://github.com/CesiumGS/cesium-omniverse/pull/541#discussion_r1388564579

Removed harcoded limit for feature id sets and texcoord sets. Imagery layers still have a maximum of 16 but the tokens are no longer hardcoded.

Instead tokens are created on demand and cached for later use.

PR #553, opened 2023-11-21
Cleanup `cesium.mdl`

lilleyse:
Merge https://github.com/CesiumGS/cesium-omniverse/pull/552 first.

Did some MDL cleanup while working on metadata.

* Parameter renames to deconflict with offset and scale properties that are coming with metadata support
  * `offset` -> `tex_coord_offset`
  * `scale` -> `tex_coord_scale`
  * `rotation` -> `tex_coord_rotation`
* Moved helper functions, exported public functions, and exported internal functions into their own sections
* Removed default parameter values from internal functions since they are always guaranteed to be set in `FabricMaterial.cpp`.
* Moved channel unpacking for feature id textures to its own function


weegeekps (approved):
LGTM

PR #554, opened 2023-11-21
Cleanup vertex attributes

lilleyse:
Merge #553 first

Cleans up vertex attribute handling in preparation for metadata support.

* Vertex attributes with `normalized: true` are no longer normalized when accessed in MDL. This behavior may change in the future, but for now it makes it much easier to implement property attributes in the follow up metadata PR.
* Moved `VertexAttributeAccessor` and `getVertexAttributeValues` functions to header to avoid having to provide explicit template instantiations.
* Added normalized types to `VertexAttributeType`. This will become more important for the metadata PR.
* Some misc cleanup in `GltfUtil.cpp`


weegeekps (approved):
LGTM

PR #555, opened 2023-11-21
Update cesium-native

lilleyse:
Update to the latest commit in cesium-native, mainly to get the fix in https://github.com/CesiumGS/cesium-native/pull/755.

### DIFF
--- a/src/bindings/PythonBindings.cpp
+++ b/src/bindings/PythonBindings.cpp
@@ -102,7 +102,7 @@ PYBIND11_MODULE(CesiumOmniversePythonBindings, m) {
         .def_readonly("percent_complete", &CesiumIonClient::Asset::percentComplete);
 
     py::class_<CesiumIonClient::Connection>(m, "Connection")
-        // Get pointer to non-static member function. May be able to use py::overload_cast<> in C++ 20
+        // Wrap non-static member function in lambda. May be able to use py::overload_cast<> in C++ 20
         .def("get_api_uri", [](CesiumIonClient::Connection& connection) { return connection.getApiUrl(); })
         .def("get_access_token", &CesiumIonClient::Connection::getAccessToken);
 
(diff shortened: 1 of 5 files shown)


src/bindings/PythonBindings.cpp
```diff
@@ -102,7 +102,8 @@ PYBIND11_MODULE(CesiumOmniversePythonBindings, m) {
         .def_readonly("percent_complete", &CesiumIonClient::Asset::percentComplete);
 
     py::class_<CesiumIonClient::Connection>(m, "Connection")
-        .def("get_api_url", &CesiumIonClient::Connection::getApiUrl)
+        // Wrap non-static member function in lambda. May be able to use py::overload_cast<> in C++ 20
+        .def("get_api_uri", [](CesiumIonClient::Connection& connection) { return connection.getApiUrl(); })
```
lilleyse, line 106: I couldn't figure out how to do this in a cleaner way. `getApiUrl` has two overloads:

```c++
  static CesiumAsync::Future<std::optional<std::string>> getApiUrl(
      const CesiumAsync::AsyncSystem& asyncSystem,
      const std::shared_ptr<CesiumAsync::IAssetAccessor>& pAssetAccessor,
      const std::string& ionUrl);
```

```c++
  const std::string& getApiUrl() const noexcept { return this->_apiUrl; }
```

First I tried `py::overload_cast<>`, which didn't work because it expects at least one template argument
Then I tried the `static_cast` approach in https://pybind11.readthedocs.io/en/stable/classes.html
I ended up just wrapping it in a lambda.

@weegeekps am I missing an obvious way to do this?

weegeekps: Seems like dynamic return types are only supported using a pattern similar to what you're already doing. I think the open question is whether or not you need to wrap `connection.getApiUrl()` in a `py::cast` call.

weegeekps (approved):
Send it.

PR #557, opened 2023-11-22
Change imagery layer dynamically in material graph

lilleyse:
Previously if you tried changing the `imagery_layer_index` in the `cesium_imagery_layer_float4` node it would have no effect. The connection between the texture node and material node was created at initialization and could never change. Feature id sets had the same problem.

This is now sort of possible. If we detect that the imagery layer changed we destroy the connection to the old texture and create a connection to the new texture, which triggers an MDL compilation and gets fixed.

Currently there are two problems with this:

* It's a bit slow. When a value changes all tiles are updated, including those in the pool. In the future we might want to update connections only when a tile becomes active. This is how CesiumJS does it for styling.
* There seems to be a bug where connections don't get updated right away. Notice how changing the feature id set from 0 to 1 works but 1 to 0 doesn't. After setting the color (which triggers the same code path) it seems to correct itself.

https://github.com/CesiumGS/cesium-omniverse/assets/915398/da19d167-826f-4bda-b5fc-56ea7b7efa0a

Even though this isn't a complete solution, it's an improvement over existing behavior and is something I need for metadata styling as well.


src/core/include/cesium/omniverse/ObjectPool.h
```diff
@@ -84,8 +84,12 @@ template <typename T> class ObjectPool {
     virtual std::shared_ptr<T> createObject(uint64_t objectId) = 0;
     virtual void setActive(std::shared_ptr<T> object, bool active) = 0;
 
+    const std::deque<std::shared_ptr<T>>& getQueue() {
+        return _queue;
+    }
+
   private:
-    std::queue<std::shared_ptr<T>> _queue;
+    std::deque<std::shared_ptr<T>> _queue;
```
corybarr, line 92: This double-ended data structure is still only being used as a queue. Is it faster for setting capacity or something?

corybarr (approved):

src/core/include/cesium/omniverse/ObjectPool.h
corybarr: Interesting. An `std::queue` is just a wrapper around `std::deque`. Then by def I imagine `std::queue` has to be at least some small amount slower.

lilleyse: The switch is actually for a different reason, `std::queue` doesn't support iterators and therefore doesn't support range based for loops which I needed to do here:

```
void FabricMaterialPool::updateShaderInput(const pxr::SdfPath& shaderPath, const pxr::TfToken& attributeName) {
    const auto& fabricMaterials = getQueue();
    for (auto& fabricMaterial : fabricMaterials) {
        fabricMaterial->updateShaderInput(
            FabricUtil::toFabricPath(shaderPath), FabricUtil::toFabricToken(attributeName));
    }
}
```

Issue #155, opened 2023-02-15
Metadata support


lilleyse:
Similar to Cesium for Unreal/Unity, add the ability to query metadata values from features in a tileset and use that to drive visualization (e.g. show/hide, materials).

PR #559, opened 2023-11-27
Add support for property textures

lilleyse:
Part of #155. Merge #558 first.

Adds support for [property textures](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_structural_metadata#property-textures) in `EXT_structural_metadata`.

image (15)

_Visualizing horizontal and vertical predicted accuracy in the Maxar Muscatatuck dataset_ - [muscatatuck.usda.zip](https://github.com/CesiumGS/cesium-omniverse/files/13479306/muscatatuck.usda.zip)

This builds off the work in #558. No changes to public-facing MDL code, but some new internal functions were added:

- `cesium_internal_property_texture_int_lookup`
- `cesium_internal_property_texture_int2_lookup`
- `cesium_internal_property_texture_int3_lookup`
- `cesium_internal_property_texture_int4_lookup`
- `cesium_internal_property_texture_normalized_int_lookup`
- `cesium_internal_property_texture_normalized_int2_lookup`
- `cesium_internal_property_texture_normalized_int3_lookup`
- `cesium_internal_property_texture_normalized_int4_lookup`

In order to keep the scope small, certain parts of `EXT_structural_metadata` are not implemented:

* Array properties are not supported
* Properties that are 16-bit per-component or larger are not supported
* Properties with components that are packed across multiple texture channels are not supported
* Properties with more than four channels are not supported

The main issue is that tiles blink in when they're loading. That's because `createConnectionsToProperties` is called in `setMaterial` instead of `initializeExistingMaterial`. This could be avoided, but it would mean that the property ids would need to be baked into the `FabricMaterialDefinition` and that feels wrong.

https://github.com/CesiumGS/cesium-omniverse/assets/915398/23036aba-d74f-4b7a-aa73-eb2020e63137


weegeekps (approved):
LGTM

PR #567, opened 2023-11-27
Add server prim

### LINKED ISSUE #560: Add prim for storing Cesium ion server metadata
In support of handling self-hosted ion servers and connections to multiple ion servers, we need move metadata specific to Cesium ion out of the `Cesium` prim and into a `CesiumIonServer` prim. This includes:

- The server url.
- The project default access token.
- The project default access token ID. This is typically hidden from the user.

weegeekps:
Resolves #560.

Adds the `CesiumIonServerPrim` definition to `CesiumUsdSchemas`. Prims of this type will contain stage-level metadata for the ion Server URL and relevant default access token information.

This also adds relationship fields to both `CesiumTilesetPrim` and `CesiumImageryPrim` for the ion server.

Creation of these prims in the stage will be handled in the next PR where I work on actually implementing usage of the relationships.


exts/cesium.usd.plugins/schemas/cesium_schemas.usda
```diff
@@ -139,6 +139,38 @@ class CesiumSessionPrim "CesiumSessionPrim" (
...
+        doc = "A string representing the token for accessing Cesium ion assets."
+    )
+
+    string cesium:projectDefaultIonAccessTokenId = "" (
+        customData = {
+            string apiName = "projectDefaultIonAccessTokenId"
+        }
+        displayName = "Project default ion Access Token ID"
+        doc = "A string representing the token ID for accessing Cesium ion assets."
+    )
```
lilleyse, line 171: Will `projectDefaultIonAccessToken` and `projectDefaultIonAccessTokenId` eventually be removed from `CesiumDataPrim`?

lilleyse (approved):

exts/cesium.usd.plugins/schemas/cesium_schemas.usda
weegeekps: Maybe. For now I think we should leave them here so people can retrieve the values they stored here.

PR #569, opened 2023-11-28
Add cesium imagery base class

corybarr:
** DO NO MERGE UNTIL FOLLOW-UP PR **

This PR creates a new prim type `CesiumIonImageryPrim` and changes the old prim type `CesiumImageryPrim` into an abstract class. This is in preparation to add a new `CesiumPolygonImageryPrim` and related classes that will handle polygon clipping and eventually polygon-based overlays. This is a breaking change to USD files.


weegeekps (commented):
Looks fine but since we're not merging this yet I'm holding off on approval.

lilleyse:
@corybarr is this ok to close since the changes are included in https://github.com/CesiumGS/cesium-omniverse/pull/599?

corybarr:
> @corybarr is this ok to close since the changes are included in #599?

@lilleyse Yes. #599 was a branch off of this, so by definition should contain all the changes.

PR #570, opened 2023-11-28
Refactored to use the new ion prims for default tokens

weegeekps:
Stepping stone PR for supporting multiple ion server prims storing default tokens.

Also contains some new stubs since I'm now using the new pieces of the schema created in my last PR.


src/core/src/UsdUtil.cpp
```diff
@@ -360,6 +367,27 @@ pxr::CesiumData getOrCreateCesiumData() {
...
+    if (isCesiumIonServer(currentIonServerPath)) {
+        auto stage = getUsdStage();
+        auto currentIonServer = pxr::CesiumIonServer::Get(stage, currentIonServerPath);
+        return currentIonServer;
+    }
+
+    return defineCesiumIonServer(currentIonServerPath);
+}
+
+[[maybe_unused]] std::vector<pxr::CesiumIonServer> getAllIonServerPrims() {
```
weegeekps, line 386: I'll be using this function very soon. It wasn't necessary for this piece.

corybarr: I know you're not using VS Code, but do you know if there's a way in it to check for functions with no references? CLion might due to the fact you added [[maybe_unused]].

PR #571, opened 2023-11-29
Added some values to CesiumIonServerPrim definition

weegeekps:
I've started to realize that in order for layers to work properly we're going to need to store as much about the ion servers in the USD stage as possible. I am probably going to store the oauth connection details in application storage but by storing the server details as prims we make the process a bit easier and easier to transfer through layers.


corybarr (approved):
Looks straightforward

PR #572, opened 2023-11-29
Cleanup in preparation for property tables

lilleyse:
Some cleanup and bug fixes while adding support for property tables.

* `DataType.h` - tried to simplify this file a bit so that `GetNativeType` is the only method for getting the C++ type. All other functions return `DataType` values, that can then be converted into native types. Also removed some of the template specialization code where more basic functions would suffice.
* Added helper functions for getting `offset`, `scale`, `noData`, `defaultValue`, `maximumValue` since this code was being repeated in a few places
* Several bug fixes which I'll point out in review comments


src/core/include/cesium/omniverse/MetadataUtil.h
```diff
@@ -292,7 +292,7 @@ void forEachStyleablePropertyTextureProperty(
             } else {
                 constexpr auto type = getTypeReverse<RawType, TransformedType>();
 
-                if constexpr (sizeof(RawType) > 1) {
+                if constexpr (getComponentByteLength<type>() > 1) {
```
lilleyse, line 295: Bug fix: get the size of the component type instead of the whole type

src/core/src/FabricMaterial.cpp
```diff
@@ -1022,6 +1070,8 @@ void FabricMaterial::reset() {
 
     _properties.clear();
 
+    destroyConnectionsToProperties();
```
lilleyse, line 1073: It's cleaner to destroy connections in `reset` and create connections in `setMaterial`

src/core/src/FabricMaterial.cpp
```diff
@@ -1388,6 +1423,9 @@ void FabricMaterial::updateShaderInput(const omni::fabric::Path& path, const omn
         attributeName == FabricTokens::inputs_feature_id_set_index) {
         destroyConnectionsToCopiedPaths();
         createConnectionsToCopiedPaths();
+    }
+
+    if (FabricUtil::isCesiumPropertyNode(attributeName)) {
         destroyConnectionsToProperties();
         createConnectionsToProperties();
     }
```
lilleyse, line 1431: Bug fix: the material wouldn't update if you changed the property id

src/core/src/FabricResourceManager.cpp
```diff
@@ -91,7 +91,10 @@ std::shared_ptr<FabricGeometry> FabricResourceManager::acquireGeometry(
 }
 
 bool useSharedMaterial(const FabricMaterialDefinition& materialDefinition) {
-    if (materialDefinition.hasBaseColorTexture() || materialDefinition.getImageryLayerCount() > 0) {
+    if (materialDefinition.hasBaseColorTexture() || materialDefinition.getImageryLayerCount() > 0 ||
+        !materialDefinition.getFeatureIdTypes().empty() ||
+        !materialDefinition.getMdlInternalPropertyAttributePropertyTypes().empty() ||
+        !materialDefinition.getMdlInternalPropertyTexturePropertyTypes().empty()) {
```
lilleyse, line 97: Bug fix: can't share materials if there are things that are unique to that material like feature ids and properties

src/core/src/FabricMaterial.cpp
weegeekps: Cleaner, probably yes, but what is the performance impact?

lilleyse: It should be the same. `reset` and `setMaterial` get called the same number of times so this is just moving the call to a different spot. And it's zero cost if the tile doesn't have metadata.

PR #573, opened 2023-11-29
Added support for property tables

lilleyse:
Part of #155. Merge #572 first.

Adds support for [property tables](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_structural_metadata#property-tables) in `EXT_structural_metadata`.

image

_Styling buildings by height_ - [buildings.zip](https://github.com/CesiumGS/cesium-omniverse/files/13505279/buildings.zip)

New internal functions were added:

- `cesium_internal_property_table_int_lookup`
- `cesium_internal_property_table_int2_lookup`
- `cesium_internal_property_table_int3_lookup`
- `cesium_internal_property_table_int4_lookup`
- `cesium_internal_property_table_normalized_int_lookup`
- `cesium_internal_property_table_normalized_int2_lookup`
- `cesium_internal_property_table_normalized_int3_lookup`
- `cesium_internal_property_table_normalized_int4_lookup`
- `cesium_internal_property_table_float_lookup`
- `cesium_internal_property_table_float2_lookup`
- `cesium_internal_property_table_float3_lookup`
- `cesium_internal_property_table_float4_lookup`

In order to keep the scope small, certain parts of `EXT_structural_metadata` are not implemented:

* Array properties are not supported
* Boolean properties are not supported
* String properties are not supported
* Matrix properties are not supported
* `UINT32`, `UINT64`, and `INT64` properties are not supported

It should be possible to support most of these in future PRs, except string properties and certain types of array properties that are difficult to implement in MDL (i.e. shader code).

Property tables values are encoded as textures and are indexed based on feature ids. Thanks to the connection system in Fabric, the output of a feature id node can be connected to the input of a property table node.

One note about texture encoding - MDL doesn't support reading from integer textures so all values are converted to float32. There could be some precision loss, e.g. for integer values larger than 2^24.


CMakeLists.txt
```diff
@@ -222,6 +222,7 @@ if(MSVC)
         /W4 # enable all warnings
         /WX # error on warnings
         /Zc:__cplusplus # make __cplusplus defined correctly
+        /bigobj
```
weegeekps, line 225: Surprised this isn't the default yet.

src/core/src/FabricPrepareRenderResources.cpp
```diff
@@ -80,6 +80,28 @@ std::vector<const CesiumGltf::ImageCesium*> getPropertyTextureImages(
     return MetadataUtil::getPropertyTextureImages(model, primitive);
 }
 
+uint64_t getPropertyTableTextureCount(
+    const FabricMesh& fabricMesh,
+    const CesiumGltf::Model& model,
+    const CesiumGltf::MeshPrimitive& primitive) {
+    if (fabricMesh.material == nullptr) {
+        return {};
```
weegeekps, line 88: NIT: probably less ambiguous to just return `0`, since the return value is a `uint64_t`.

weegeekps (approved):
Looks good and works well. One minor nitpicky comment but it can be fixed in a future PR or not at all.

src/core/src/FabricPrepareRenderResources.cpp
lilleyse: Thanks for catching this, it wasn't intentional, just a copy paste mistake.

lilleyse: https://github.com/CesiumGS/cesium-omniverse/pull/573/commits/6e74948cd616e2edeaeee40b147f99c686b72f66

PR #576, opened 2023-11-30
Fix tile flickering when doing styling

lilleyse:
There was a bug mentioned in https://github.com/CesiumGS/cesium-omniverse/pull/559 that started to feel more and more important to address.

> The main issue is that tiles blink in when they're loading. That's because `createConnectionsToProperties` is called in `setMaterial` instead of `initializeExistingMaterial`. This could be avoided, but it would mean that the property ids would need to be baked into the `FabricMaterialDefinition` and that feels wrong.

I implemented the solution described there. `FabricMaterialDefinition` now has a list of `PropertyDefinition` which includes the property id, property type, and feature id set index (if applicable). Feature ID nodes are connected to property table nodes in `initializeNodes` and internal/external property nodes are connected in `initializeExistingMaterial`. Since this all happens during material initialization it means the shader doesn't need to be compiled dynamically, which is what led to the flickering before. This also made certain parts of the code cleaner. It also probably improves performance.

The downside is that tiles can't use the same material pool unless they have the same property ids. In practice this is probably ok since most tiles within a tileset have the same property ids.

Before:

https://github.com/CesiumGS/cesium-omniverse/assets/915398/23036aba-d74f-4b7a-aa73-eb2020e63137

After:

https://github.com/CesiumGS/cesium-omniverse/assets/915398/d4a95e81-843a-4695-9fdc-0e960eb40cdb

Tested with [muscatatuck.usda.zip](https://github.com/CesiumGS/cesium-omniverse/files/13479306/muscatatuck.usda.zip)


corybarr (approved):
This makes sense. Looks clean.

lilleyse:
I squeezed in a separate bug fix where tiles weren't updating when changing the `propertyId`: https://github.com/CesiumGS/cesium-omniverse/pull/576/commits/733b3469ea59fb1b13b45b62e32ca5647d20a1e8

PR #579, opened 2023-12-01
Reduce number of metadata warnings

lilleyse:
Currently a warning is printed every time an unsupported metadata property is encountered. This happens for every tile in the tileset, so the number of warnings can add up.

This PR skips property processing unless you're using a tileset material. This should help when loading OSM buildings.

I have a more robust fix in mind, something like CesiumJS's `oneTimeWarning`, but I don't want to hold up the release any longer.


corybarr (approved):
Looks good. Feels a bit more like R to me with the metaprogramming than what I usually see in C++.

PR #590, opened 2023-12-14
Update cesium-native to 0.31.0

lilleyse:


r-veenstra (approved):
Changes all appear fine and confirming data loads

Issue #562, opened 2023-11-27
UI for selecting current ion server


weegeekps:
To support multiple ion server connections, we need a selector which allows the user to choose which ion server they want to use for Quick Add and adding tilesets and imagery from the Asset Window.

PR #578, opened 2023-12-01
Implement support for multiple ion servers

weegeekps:
Resolves #561. 
Resolves #562. 
Resolves #563. 
Resolves #565. 
Resolves #566.

Finishes the implementation of multiple ion server support for self-hosted ion.

Issue #565, opened 2023-11-27
Update Asset Window to support multiple ion servers


weegeekps:
The Asset Window needs to support multiple ion servers. It should have the same selector that's on the Main Window.

Issue #563, opened 2023-11-27
Update token selector to support multiple ion servers


weegeekps:
The token selector needs to be updated to support multiple ion servers. Which server is updated is based on the selected session server.

PR #591, opened 2023-12-15
Add support for multiple ion servers in UI

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/562.

Follow up to https://github.com/CesiumGS/cesium-omniverse/pull/578 that adds a cesium ion server selector in the UI. There are still a few UI components that are not addressed in this PR but will come in a follow up:

* https://github.com/CesiumGS/cesium-omniverse/issues/563
* https://github.com/CesiumGS/cesium-omniverse/issues/565
* Dynamically populating the Quick Add assets based on the server

In addition, there are some bugs related to saving and loading:

* Saving a .usda with multiple servers for the first time causes a server disconnect
* Loading a .usda with multiple servers doesn't resume the connection

Projects with the single default ion server don't seem to have these issues, which is good.

Also...

* Creating a new project always requires connecting to ion. That's because there's no existing connection to resume. I think this is expected behavior, but would have to verify with Cesium for Unreal.

Finally, while I started to implement something that would query the ion api url automatically, this didn't really work with the existing code. @r-veenstra any tutorial content will need to be updated so that the api url is filled in instead of left blank. E.g. for `https://self-hosted.cesium.com/` it would be `https://api.self-hosted.cesium.com/`.

Obviously there are a lot of improvements to make here still, but it's certainly usable. Here I add CWT from ion.cesium.com and GTOPO from self-hosted.cesium.com.

https://github.com/CesiumGS/cesium-omniverse/assets/915398/e4216afa-c9aa-4800-b175-c6210bd10386


r-veenstra (approved):
Okay, this all appears to be working in my testing, with exception to the bugs and limitations outlined in the PR.

I have noticed one additional issue where the "Create a new token" option in the Token UI doesn't appear to create tokens with self hosted, but works ok with regular ion.

lilleyse:
> Creating a new project always requires connecting to ion. That's because there's no existing connection to resume. I think this is expected behavior, but would have to verify with Cesium for Unreal.

This is fixed in https://github.com/CesiumGS/cesium-omniverse/pull/594. It should no longer sign you out each time. What was happening was that the property change notifications [here](https://github.com/CesiumGS/cesium-omniverse/blob/a2dc0ef28ce3be30a99272418ccdeb6217351b20/exts/cesium.omniverse/cesium/omniverse/extension.py#L463-L466) were triggering a disconnect [here](https://github.com/CesiumGS/cesium-omniverse/blob/a2dc0ef28ce3be30a99272418ccdeb6217351b20/src/core/src/Context.cpp#L435-L436). Since property change notifications are now ignored after prim add notifications, this no longer happens.

PR #592, opened 2023-12-15
Add backwards compatibility for older projects created before CesiumIonServerPrim

lilleyse:
* Transfers the default access token from the `/Cesium` prim to the `/CesiumServers/IonOfficial` prim
* Binds `/CesiumServers/IonOfficial` to any tilesets and imagery that don't already have server bindings on stage load


r-veenstra (approved):
The code makes sense and I can confirm all the Cesium for Omniverse sample examples using the old schema have loaded successfully in USD Composer.

Issue #483, opened 2023-10-04
Perform refactoring pass on UsdNotificationHandler


weegeekps:
The `UsdNotificationHandler` needs a refactoring pass performed to clean it up a bit and make the code a bit more self documenting. There are two primary issues that need to be resolved, outlined below, but whoever performs this work should use judgement to decide if there are other improvements that could be made.

1. It's probably time to create a container class for `ChangedPrim`s rather than using a pure vector. We have several cases where we are just passing in boilerplate parameters: in particular the `propertyName` parameter is only used by changed prims rather than added or removed prims. Abstracting adding entries to the collection would allow for self-documenting improvements, ergonomic improvements, and simplify where we need to handle logging.
2. We are also creating a fair number of unnecessary temporary objects in our usages of `emplace_back` that we should clean up. [Clang-tidy is flagging these](https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-emplace.html) and the reasoning is [outlined best by this article](https://andreasfertig.blog/2023/04/push_back-vs-emplace_back-when-to-use-what/). Effectively the way we're using `emplace_back` right now doesn't fully provide us with all the benefits and is using more cycles than what is necessary.

lilleyse:
For point 2 -- C++20 added support for `emplace_back` for struct member initialization. Once we're able to switch to C++20 we should make that change. This affects a few areas of the code, not just `UsdNotificationHandler`.

[...]

PR #594, opened 2023-12-15
Clean up UsdNotificationHandler

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/483

Previously `UsdNotificationHandler` would poll notifications and `Context` would process them each frame. This PR moves the processing into `UsdNotificationHandler` to help simplify `Context`.

There are some other performance / stability improvements:

* Ignore property change notifications that occur immediately after prim added notifications
* Consolidate property changes so that they can be processed together

This should help prevent assets from going into an invalid state. It should also help improve performance.

In terms of https://github.com/CesiumGS/cesium-omniverse/issues/483:

* I didn't add a container class for `ChangedPrim` but I simplified how prims are added to the vector. Personally I'm ok with the current setup. A vector is a simple way to ensure that notifications are processed in the order they're added.
* I added some inline comments about improvements once we can switch to C++ 20.

### DIFF
--- a/src/core/src/Context.cpp
+++ b/src/core/src/Context.cpp
@@ -113,6 +113,10 @@ std::shared_ptr<TaskProcessor> Context::getTaskProcessor() {
     return _taskProcessor;
 }
 
+std::shared_ptr<CesiumAsync::AsyncSystem> Context::getAsyncSystem() {
+    return _asyncSystem;
+}
+
 std::shared_ptr<HttpAssetAccessor> Context::getHttpAssetAccessor() {
     return _httpAssetAccessor;
 }
@@ -157,6 +161,8 @@ void Context::clearStage() {
 void Context::reloadStage() {
     clearStage();
 
+    _usdNotificationHandler.onStageLoaded();
+
     auto& fabricResourceManager = FabricResourceManager::getInstance();
     fabricResourceManager.setDisableMaterials(getDebugDisableMaterials());
     fabricResourceManager.setDisableTextures(getDebugDisableTextures());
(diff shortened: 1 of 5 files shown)


lilleyse:
This PR also fixes a sign-out issue. See https://github.com/CesiumGS/cesium-omniverse/pull/591#issuecomment-1858381563 for details.

src/core/src/Context.cpp
```diff
@@ -168,60 +174,31 @@ void Context::reloadStage() {
...
-        const auto& path = prim.GetPath();
-        if (UsdUtil::isCesiumIonServer(path)) {
-            SessionRegistry::getInstance().addSession(*_asyncSystem, _httpAssetAccessor, path);
+    // For backwards compatibility. Add default ion server to tilesets without a server.
+    const auto& tilesets = AssetRegistry::getInstance().getAllTilesets();
+    for (const auto& tileset : tilesets) {
+        const auto ionServerPath = tileset->getIonServerPath();
+        if (ionServerPath.IsEmpty()) {
+            const auto tilesetPrim = UsdUtil::getCesiumTileset(tileset->getPath());
+            tilesetPrim.GetIonServerBindingRel().SetTargets({defaultIonServerPath});
```
corybarr, line 185: Is it worth notifying the users the file is being altered? Our customers and OV users in general are pretty early adopters, so it's arguable that a release not vs something in-app is sufficient.

Also, should there be an issue to remove this in some number of months in the future?

corybarr (approved):
Looks good to me. More as a note to myself, the CesiumIonImageryPrim will need to support the Ion server. Related to this, I'm not sure "Ion" is accurately descriptive with self-hosting being an option now.

src/core/src/Context.cpp
```diff
@@ -157,6 +161,8 @@ void Context::clearStage() {
 void Context::reloadStage() {
     clearStage();
 
+    _usdNotificationHandler.onStageLoaded();
```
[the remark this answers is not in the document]
lilleyse: I think I still prefer Context controlling the execution order.

src/core/src/Context.cpp
corybarr: Since we're pre-release, IMO a console message seems sufficient. v1.0 seems like a good time to remove the backwards compatibility.

corybarr:
> > Looks good to me. More as a note to myself, the CesiumIonImageryPrim will need to support the Ion server. Related to this, I'm not sure "Ion" is accurately descriptive with self-hosting being an option now.
> 
> "Ion" is fine. ion self-hosted and ion saas are both considered ion servers.

Ah, you're right. In my first weeks here, the name "ion" registered as a service hosted by Cesium. I need to refine that. It's still Ion if  self-hosted.

src/core/src/Context.cpp
lilleyse: Added warning in https://github.com/CesiumGS/cesium-omniverse/pull/594/commits/fb181ccbab57d371da9a5a9cc9e6b10739ada315

lilleyse: Issue here: https://github.com/CesiumGS/cesium-omniverse/issues/598

PR #596, opened 2023-12-15
Clean up AssetRegistry

lilleyse:
Merge https://github.com/CesiumGS/cesium-omniverse/pull/594 first

* Removed `get_all_tileset_paths` / `getAllTilesetPaths` from Python binding layer since the tileset paths can be queried from the USD stage (there's now a helper in `usdUtil.py`)
* Return `std::shared_ptr` instead of `std::optional<std::shared_ptr>` since `nullptr` is a good enough replacement for `std::nullopt`.

This is setting the groundwork for a larger `AssetRegistry` change in a future PR.


corybarr (approved):
This removal of `std::optional` and `std::nullopt` makes the code more direct and readable, in particular for those of us who've been living in language other than C++ for years.

PR #601, opened 2023-12-22
Fix pulling kit-sdk dependencies

lilleyse:
One of the old nvidia build tools we were using (`repo_repoman 0.1.1-beta2`) was removed from the nvidia server and was causing clean builds to fail. This PR updates `packman` to the same version as https://github.com/NVIDIA-Omniverse/kit-extension-template-cpp and removes `repoman` since we don't actually need it. `install.py` now just pulls the dependencies with packman.


extern/nvidia/tools/packman/bootstrap/configure.bat
```diff
@@ -95,11 +95,16 @@ if exist "%PM_PYTHON%" (
...
 
-:: Perform atomic rename
-rename "%TEMP_FOLDER_NAME%" "%PM_PYTHON_VERSION%" 1> nul
-:: Failure during move, need to clean up and abort
+:: Perform atomic move (allowing ovewrite, /y)
+move /y "%TEMP_FOLDER_NAME%" "%PM_PYTHON_DIR%" 1> nul
+:: Verify that python.exe is now where we expect
+if exist "%PM_PYTHON%" goto PACKMAN
+
+:: Wait a second and try again (can help with access denied weirdness)
```
corybarr, line 103: Just curious is this is from experience or purely defensive

corybarr (approved):
This all looks good to me. Our build system is the thing I understand the least, so I'm not super clean on what scripts we own and what are used by packman and the OV build scripts.

extern/nvidia/tools/packman/bootstrap/configure.bat
lilleyse: I have no idea. I just copied the packman files from https://github.com/NVIDIA-Omniverse/kit-extension-template-cpp

PR #604, opened 2023-12-28
Update cesium.usd.plugins version

lilleyse:
Updates the version number for `cesium.usd.plugins`. We should be doing this any time we make changes to the USD schemas otherwise `cesium.omniverse` may reference an incompatible version of `cesium.usd.plugins`. This is more likely to happen now that the extensions are published as community extensions. The release guide has been updated as well.


r-veenstra (approved):
Wokring well here and instruction updates make sense

PR #605, opened 2023-12-28
Remove retain paths workaround

lilleyse:
The workaround in https://github.com/CesiumGS/cesium-omniverse/pull/442 is no longer needed. The underlying issue was fixed in `105.1+release.127680.dd92291b.tc` which we've been using for a few months now.

I confirmed that adding CWT + Bing, removing CWT + Bing, adding CWT + Bing doesn't trigger a crash.


r-veenstra (approved):
Confirming adding CWT + Bing, removing CWT + Bing, adding CWT + Bing doesn't trigger a crash.

Issue #441, opened 2023-08-17
Upgrade to clang 15


lilleyse:
Originally suggested by @corybarr:

We get IntelliSense errors on Windows via `glm.hpp` via `yvals_core.h` which are fixed by upgrading to a newer version of clang.

![image (2)](https://github.com/CesiumGS/cesium-omniverse/assets/915398/e4ae4eca-02c7-405a-912b-587aebd2e636)

Also we want to update the download link for LLVM to https://releases.llvm.org/ instead of https://llvm.org/builds/. (The snapshot builds are no longer updated. Use the [regular releases](https://llvm.org/releases/) instead.)

### Steps

There's a few occurrences of `14` that will have to change to `15` in these files:

* `.github/workflows/main.yml`
* `.vscode/tasks.json`
* `docs/developer-setup/README.md`

lilleyse:
Right now it's easier to switch to clang-15 since clang-16 isn't pre-installed in the GitHub Action `ubuntu-latest` runner. Updated the title of this issue to reflect that.

PR #611, opened 2024-01-08
Upgrade to clang-15

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/441

I wanted to experiment with some ways to speed up lldb debugging that required clang-15. Unfortunately those experiments didn't work, but it was still worth opening this PR.

For posterity, this is what I was trying to do in `launch.linux.json`:

```
"type": "lldb",
"initCommands": [
  "settings set symbols.enable-lldb-index-cache true",
  "settings set symbols.load-on-demand true"
]
```

Based on this thread: https://discourse.llvm.org/t/lldb-startup-takes-15-seconds-parsing-symbol-tables/65487/5

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -115,7 +115,7 @@ jobs:
         if: ${{ runner.os == 'Linux' && matrix.config.name != 'AlmaLinux 8 - GCC' }}
         run: |
           sudo apt update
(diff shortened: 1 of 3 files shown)


corybarr (approved):
Very straightforward

PR #621, opened 2024-01-17
Update to Kit SDK 105.1.2

lilleyse:
Update to Kit SDK 105.1.2 which is used by USD Composer 2023.2.2.


r-veenstra:
@lilleyse tested this build in the following successfully

r-veenstra (approved):

PR #622, opened 2024-01-18
Use default GitHub Actions runners

lilleyse:
See https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/

We no longer need to use `windows-latest-l` or `ubuntu-latest-m`


lilleyse:
Thanks for the tip @kring: https://github.com/CesiumGS/cesium-unreal/pull/1323#issuecomment-1897838435

corybarr (approved):

PR #624, opened 2024-01-18
Change conan version in developer setup README

timoore:
conan version 1.58 was causing a an error with the libtool recipe, which is being sucked in from some unknown dependency. It's easiest for the moment to move on to 1.60 as required by libtool.

Also, suggest "update-alternatives" as an alternative way to manage clang-15 symlinks.


lilleyse (approved):

lilleyse:
@timoore as a follow up to this could you replace other occurrences of `1.58.0` to `1.60.0` and publish a new docker image?
 
See https://github.com/CesiumGS/cesium-omniverse/blob/main/docs/release-guide/push-docker-image.md

Issue #616, opened 2024-01-09
Make FabricTexture generically support different channels and bytes per channel


corybarr:
Sean's comment that I agree with: "I don't really like how imagery overlay concepts are leaking into this class. I think FabricTexture just needs to be able to generically support different channels and bytesPerChannel that could be present in ImageCesium, and that would cover the bases for imagery."

Issue #600, opened 2023-12-21
Globe anchor position out of sync after programmatic creation


r-veenstra:
Performing some testing in main post https://github.com/CesiumGS/cesium-omniverse/pull/594.  

The globe anchor script featured in https://cesium.com/learn/omniverse/omniverse-placing-objects/ no longer works as expected.  Working fine in https://github.com/CesiumGS/cesium-omniverse/commit/a2dc0ef28ce3be30a99272418ccdeb6217351b20, broken as of https://github.com/CesiumGS/cesium-omniverse/commit/bd9d2d25a77e596f75d64b37fe247a8e58fa01a5

Script copied below for records.

```
import omni.kit.commands
import omni.usd
from pxr import Sdf, UsdGeom, Usd
from cesium.omniverse.api.globe_anchor import anchor_xform_at_path
from asyncio import ensure_future

trees = [
	{"name":"tree_01","latitude":37.79130, "longitude":-122.39032,"height":-28.88791},
	{"name":"tree_02","latitude":37.79114, "longitude":-122.39018,"height":-28.88791},
	{"name":"tree_03","latitude":37.79147, "longitude":-122.39042,"height":-28.88791},
	{"name":"tree_04","latitude":37.79167, "longitude":-122.39053,"height":-28.88791},
]

context = omni.usd.get_context()
stage = context.get_stage()

async def add_trees():
	for t in trees:
	
		# Create the prims name
		payload_prim_path = "/World/"+t["name"]

		# Create the prim for the payload
		payload_prim: Usd.Prim = UsdGeom.Xform.Define(stage, Sdf.Path(payload_prim_path)).GetPrim()
		
		# Add the tree payload to the prim
		omni.kit.commands.execute("AddPayload",
			stage=stage,
			prim_path = payload_prim_path,
			payload=Sdf.Payload(
				assetPath = "http://omniverse-content-production.s3-us-west-2.amazonaws.com/Assets/Vegetation/Plant_Tropical/Cuban_Royal_Palm.usd",
				primPath = Sdf.Path.emptyPath
			)
		)

		# Add a Cesium Globe Anchor
		anchor_xform_at_path(payload_prim_path)
		
		# Wait a frame to allow notifications to complete
		await omni.kit.app.get_app().next_update_async()
		
		# Set the latitude, longitude and height for the tree
		payload_prim.GetAttribute("cesium:anchor:geographicCoordinates").Set(
			(t["latitude"], t["longitude"], t["height"])
[...]

Issue #588, opened 2023-12-14
Excessive property warnings when using custom material and property lookups


r-veenstra:
Using a custom material to access data in property tables triggers excessive warnings when using property heavy datasets such as Cesium OSM Buildings.

Loading the below usda produces approximately 75,000 warnings.

[osm_buildings_with_properties.zip](https://github.com/CesiumGS/cesium-omniverse/files/13670945/osm_buildings_with_properties.zip)

It may be better to only display warnings for properties that the material is trying to use, as opposed to every property in the tileset.

![image](https://github.com/CesiumGS/cesium-omniverse/assets/123468416/0e14834d-5fee-4f1a-8179-8326f29a940a)

lilleyse:
Yup, I'd like to fix this soon. https://github.com/CesiumGS/cesium-omniverse/pull/579 was the first step, but I have a better fix in mind that I haven't gotten to yet.

r-veenstra:
I still seem to be getting this, loading https://github.com/CesiumGS/cesium-omniverse-samples/blob/main/tutorials/07_style_by_properties/07_style_by_properties.usda when running a build of 1d4f98be7001c044a8d4b39759f50895e40a6738 which should have the latest #626 codebase

![image](https://github.com/CesiumGS/cesium-omniverse/assets/123468416/0abec784-e774-435f-9ee6-7653cb949ac3)

Issue #584, opened 2023-12-04
Adding a Globe Anchor to a scaled prim changes size or position in the stage


corybarr:
To reproduce:

1. Create a prim. Do not alter the scale. Add a globe anchor. He prim remains unaltered.
2. Create another prim. Alter the scale and /or translation. Add a globe anchor. The scale and/or translation will change depending on the prim's pre-anchored transformations.

It's possible to avoid the issue, but the user has to set the translation first, add the globe anchor, then update the scale.

lilleyse:
I confirmed that this is fixed in https://github.com/CesiumGS/cesium-omniverse/pull/626.

Existing transforms are unaltered when the globe anchor is added, and you can safely update them afterwards.

PR #626, opened 2024-01-18
Large refactor of the C++ code

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/539
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/584
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/600
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/588
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/616

This is a wide reaching cleanup PR that has both functional changes and C++ style changes. Ideally this would have been spit into multiple PRs, but here we are 🤕 

#### Globe Anchors

* The Python API for adding globe anchors has been removed. The `Python <-> C++` syncing issues have been fixed and globe anchors can now be defined directly in USD at the Python layer.
* Globe anchors no longer add a `cesium:transform` op to the attached prim. Instead they modify the `translate`, `rotate` (all variants), and `scale` ops directly.
* Several bugs were fixed
  * https://github.com/CesiumGS/cesium-omniverse/issues/539
  * https://github.com/CesiumGS/cesium-omniverse/issues/584
  * https://github.com/CesiumGS/cesium-omniverse/issues/600
* The naming of transforms is now more consistent throughout the code

#### Asset Registry

* Every Cesium prim type now has a corresponding `Omni` class. All `Omni` objects are stored in and accessed from `AssetRegistry`. A lot of time was spent making sure `UsdNotificationHandler` was accurately updating `AssetRegistry` and the various `Omni` classes.
* `OmniPolygonImagery` and `OmniIonImagery` now own the underlying cesium-native raster overlay object. This will allow multiple tilesets to use the same imagery resource once #585 is implemented.
* `GlobeAnchorRegistry` and `SessonRegistry` have been consolidated into `AssetRegistry`

#### Context

* Most singletons and global variables have been removed. Anything that was global state is now stored in `Context`.
* Multiple `Context` objects can now be created, if desired, such as in unit tests
* Most ion related code has moved into `CesiumIonServerManager`. I hoped to do more refactoring in this PR but ran out of time.

#### File changes

* Created `CppUtil` for various C++ or standard library related helper functions
* Renamed `Fabric***Definition` to `Fabric***Descriptor`
* Created separate files for many of the `Fabric` structs

#### Style changes

* `++i` instead of `i++` for the slight performance gain and to match cesium-native conventions
* Use `const` more frequently
* Helper functions in `***Util.h` files are now scoped in sub namespaces of `cesium::omniverse`
* Add `p` prefix to pointers, similar to cesium-native style
* More aggressively deleted copy constructors and copy assignment operators to prevent coding errors
* Consistently using brace initialization for member variables in header files
* Renamed `srw` and `isrw` to `fabricStage` and `iFabricStage`
* More aggressive about using forward declarations in header files
* `std::string_view` instead of `const char*` in certain places
* ~`pxr::` -> `PXR_NS::` which is more consistent with other projects like `kit-extension-template-cpp`~
* Tried to fix most errors reported by `clang-tidy` but there are still a few remaining

#### Other changes

* The code is much more resilient if `/Cesium` or `/CesiumGeoreference` prims aren't defined or `rel`'s are empty
* Added ability to log `oneTimeWarning`, which fixes [#588](https://github.com/CesiumGS/cesium-omniverse/issues/588)
* Updated `FabricTexture` to handle more channels and bit depths, which fixes [#616](https://github.com/CesiumGS/cesium-omniverse/issues/616)
* `CesiumOmniverse.h` no longer depends on `GfMatrix4d`. This is one step closer towards towards https://github.com/CesiumGS/cesium-omniverse/issues/3
* Reduced the amount of `Usd::Gf` math. `glm` is used almost everywhere now.


timoore:
When I write a mega-change like this, I try to break it up after the fact into commits that could at least compile and might even work. I'm a fan of the workflow described in https://git-scm.com/docs/git-stash#Documentation/git-stash.txt-Testingpartialcommits
for going about this. I do this for the obvious reasons of:
* Telling a story about the design and progression of the changes via the commits;
* Providing a better surface for `git bisect` if that is needed.

I have no idea if it is practical to do this for this commit, or even a good use of time, so this is just a noob suggestion!

lilleyse:
Thanks for the suggestion @timoore. It could be tricky to detangle things at this point, but I agree with everything you said. I haven't tried that specific workflow but I'll give it a shot the next time this happens (which hopefully won't be very often).

timoore (commented):
Drive-by review

src/core/include/cesium/omniverse/CppUtil.h
```diff
@@ -0,0 +1,104 @@
...
+#include <cstdint>
+#include <functional>
+#include <optional>
+#include <set>
+#include <unordered_set>
+#include <vector>
+
+namespace cesium::omniverse::CppUtil {
+
+template <typename T, typename L, uint64_t... I> const auto& dispatchImpl(std::index_sequence<I...>, L lambda) {
```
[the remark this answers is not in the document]
lilleyse: I borrowed this solution from stack overflow, but I can't seem to find the post right now...

This function is called by `CALL_TEMPLATED_FUNCTION_WITH_RUNTIME_DATA_TYPE` and related macros. When used in that context, `dispatchImpl` returns an array storing function pointers for all enumerations, and `dispatch` calls the function for the given enumeration.

I had tried to switch to `std::array` before, and thought I was using `sizeof...(I)` correctly, but got blocked by something. I just tried it again and it worked🤷‍♂️. I pushed the change: https://github.com/CesiumGS/cesium-omniverse/pull/626/commits/61f2698a8bfd93c8d3fe51778b90fbde11049eff

I did just try `auto&& lambda`, but that only seems possible with C++ 20.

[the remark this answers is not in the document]
timoore: > This function is called by `CALL_TEMPLATED_FUNCTION_WITH_RUNTIME_DATA_TYPE` and related macros. When used in that context, `dispatchImpl` returns an array storing function pointers for all enumerations, and `dispatch` calls the function for the given enumeration.

Ah, OK. I searched for uses of dispatch() but didn't realize that dispatchImpl is actually the one used.
> 
> I had tried to switch to `std::array` before, and thought I was using `sizeof...(I)` correctly, but got blocked by something. I just tried it again and it worked🤷‍♂️. I pushed the change: [61f2698](https://github.com/CesiumGS/cesium-omniverse/commit/61f2698a8bfd93c8d3fe51778b90fbde11049eff)
> 
> I did just try `auto&& lambda`, but that only seems possible with C++ 20.

My bad, that only works in lambda argument lists pre C++20.

src/core/include/cesium/omniverse/FabricMaterialInfo.h
```diff
@@ -0,0 +1,58 @@
+#pragma once
+
+#include "cesium/omniverse/FabricTextureInfo.h"
+
+#include <glm/glm.hpp>
+
+#include <optional>
+
+#ifdef CESIUM_OMNI_MSVC
```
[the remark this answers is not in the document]
lilleyse: I opened a cesium-native PR for that: https://github.com/CesiumGS/cesium-native/pull/796

exts/cesium.omniverse/cesium/omniverse/usdUtils/usdUtils.py
```diff
@@ -42,75 +74,84 @@ def add_tileset_ion(name: str, tileset_id: int, token: str = "") -> str:
...
-    assert imagery.GetPrim().IsValid()
-    parent = imagery.GetPrim().GetParent()
-    assert parent.IsA(CesiumTileset)
 
     imagery.GetIonAssetIdAttr().Set(asset_id)
     imagery.GetIonAccessTokenAttr().Set(token)
 
     server_prim_path = get_path_to_current_ion_server()
-    if server_prim_path != Sdf.Path.emptyPath:
+    if server_prim_path != "":
```
corybarr, line 90: In general, we're treating prim paths as strings. What's the motivation for using strings over the `Sdf.Path` class?

exts/cesium.omniverse/cesium/omniverse/ui/main_window.py
```diff
@@ -132,8 +132,6 @@ def _setup_subscriptions(self):
         )
 
     def _on_update_frame(self, _e: carb.events.IEvent):
-        self._cesium_omniverse_interface.on_update_ui()
```
corybarr, line 135: Is this OK to delete? This calls this C++ code:

```
void Context::onUpdateUi() {
    // A lot of UI code will end up calling the session prior to us actually having a stage. The user won't see this
    // but some major segfaults will occur without this check.
    if (!UsdUtil::hasStage()) {
        return;
    }
```

exts/cesium.usd.plugins/cesium/usd/plugins/CesiumUsdSchemas/__init__.pyi
```diff
@@ -6,19 +6,15 @@ import pxr.UsdGeom
 
 __MFB_FULL_PACKAGE_NAME: str
 
-class CartographicPolygon(pxr.UsdGeom.Gprim):
+class CartographicPolygon(pxr.UsdGeom.BasisCurves):
```
corybarr, line 9: This should already be in the main branch. Are the `__init__.py` files manually maintained?

I just checked out main and ran `regenerate_schema.bat` followed by `genStubs.sh`. This file wasn't created, but I have errors in `genStubs.sh`

src/core/include/cesium/omniverse/AssetRegistry.h
```diff
@@ -1,65 +1,109 @@
...
+    [[nodiscard]] OmniData* getFirstData() const;
+
+    OmniTileset& addTileset(const PXR_NS::SdfPath& path);
+    void removeTileset(const PXR_NS::SdfPath& path);
+    [[nodiscard]] OmniTileset* getTileset(const PXR_NS::SdfPath& path) const;
+    [[nodiscard]] const std::vector<std::unique_ptr<OmniTileset>>& getAllTilesets() const;
+
+    OmniIonImagery& addIonImagery(const PXR_NS::SdfPath& path);
+    void removeIonImagery(const PXR_NS::SdfPath& path);
+    [[nodiscard]] OmniIonImagery* getIonImagery(const PXR_NS::SdfPath& path) const;
```
corybarr, line 58: What advantages are being gained by switching from smart to raw pointers?

src/core/src/UsdNotificationHandler.cpp
```diff
@@ -83,300 +38,453 @@ bool inSubtree(const pxr::SdfPath& path, const pxr::SdfPath& descendantPath) {
...
+    }
+
+    if (updateBindings) {
+        updateGlobeAnchorBindings(pContext, globeAnchorPath);
     }
 }
 
-void processCesiumIonServerChanged(const pxr::SdfPath& ionServerPath, const std::vector<pxr::TfToken>& properties) {
+void processCesiumIonServerChanged(
+    Context* pContext,
```
corybarr, line 422: Pretty minor, but to use `const` more frequently this could be

```suggestion
    Context* const pContext,
```

corybarr:
This isn't related to this PR, but it is making me ask the question of if there's a better approach than having a `CesiumCartographicPolygon` contain one `BasisCurves` and one `GlobeAnchor`. I'm wondering if we want to have `CesiumCartographicPolygon` prims potentially share a `GlobeAnchor`. If we ever wanted to shift all `CartographicPolygons`' `GlobeAnchors`, we'd need to do each individually. Maybe that's rare in workflows, or could be something in PowerTools.

src/core/src/AssetRegistry.cpp
```diff
@@ -1,108 +1,286 @@
...
     }
 
     return nullptr;
 }
 
-std::shared_ptr<OmniPolygonImagery> AssetRegistry::getPolygonImagery(const pxr::SdfPath& path) const {
-    for (const auto& imagery : _polygonImageries) {
-        if (imagery->getPath() == path) {
-            return imagery;
+const std::vector<std::unique_ptr<OmniPolygonImagery>>& AssetRegistry::getAllPolygonImageries() const {
```
corybarr, line 129: Kind of a nit and coming from C# where properties/getter/setter names are strictly defined, I wonder if it makes sense to drop change this to `getPolygonImageries()`.  The `All` seems redundant if the names is plural and returns a container. (I can swap this in my PR--no need here.)

exts/cesium.omniverse/cesium/omniverse/usdUtils/usdUtils.py
lilleyse: I think we should switch everything over to paths eventually. For now I just wanted to be consistent with using strings.

exts/cesium.omniverse/cesium/omniverse/ui/main_window.py
lilleyse: It should be ok to delete, but I need to test it some more. `onUpdateUi` was removed and `onUpdateFrame` now ticks the servers.

exts/cesium.usd.plugins/cesium/usd/plugins/CesiumUsdSchemas/__init__.pyi
lilleyse: These changes came from me running `genStubs.bat`, no manual edits. Are you able to run the `.bat` version?

src/core/src/UsdNotificationHandler.cpp
lilleyse: In general we don't really use const when passing by value (in this case the pointer is being passed by value).

However there could be a case for this being `const Context* pContext`.

src/core/src/AssetRegistry.cpp
lilleyse: Good point, and I can make the change here.

src/core/src/UsdNotificationHandler.cpp
lilleyse: > However there could be a case for this being `const Context* pContext`.

Never mind, it needs be non-const in this function

src/core/include/cesium/omniverse/Context.h
```diff
@@ -1,166 +1,93 @@
...
-#include <atomic>
 #include <filesystem>
 #include <memory>
 #include <vector>
 
+#include <gsl/span>
+
+namespace omni::fabric {
+class StageReaderWriter;
+}
```
corybarr, line 13: Super nit. For consistency, you might want a space after this line. I'm surprised clang-format doesn't add it.

src/core/include/cesium/omniverse/FabricAttributesBuilder.h
```diff
@@ -1,17 +1,22 @@
...
+class StageReaderWriter;
+}
 
 namespace cesium::omniverse {
 
 class FabricAttributesBuilder {
   public:
     void addAttribute(const omni::fabric::Type& type, const omni::fabric::Token& name);
-    void createAttributes(const omni::fabric::Path& path) const;
+    void createAttributes(omni::fabric::StageReaderWriter& fabricStage, const omni::fabric::Path& path) const;
```
corybarr, line 14: Are there situations where we need to write to a specifically provided stage? I'm wondering if having to obtain and pass in the stage creates more work when creating attributes Fabric prims.

src/core/include/cesium/omniverse/FabricMaterial.h
```diff
@@ -81,6 +87,10 @@ class FabricMaterial {
         const std::vector<std::pair<omni::fabric::Type, omni::fabric::Token>>& additionalAttributes = {});
     void createTexture(const omni::fabric::Path& path);
     void createImageryLayer(const omni::fabric::Path& path);
+    void createImageryLayerResolverCommon(
```
corybarr, line 90: Thanks. This was on my TODO.

src/core/src/Context.cpp
```diff
@@ -1,39 +1,31 @@
...
-#include <CesiumUsdSchemas/globeAnchorAPI.h>
-
 #ifdef CESIUM_OMNI_MSVC
 #pragma push_macro("OPAQUE")
 #undef OPAQUE
 #endif
 
 #include <Cesium3DTilesContent/registerAllTileContentTypes.h>
-#include <Cesium3DTilesSelection/Tileset.h>
 #include <CesiumUsdSchemas/data.h>
```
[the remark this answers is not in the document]
lilleyse: Yeah, I just removed it. I haven't gone through the .cpp files to remove unused headers yet. Not sure if I'll have the energy to do it in this PR.

src/core/include/cesium/omniverse/FabricAttributesBuilder.h
lilleyse: This class should probably follow what the other classes are doing at take a `pContext` in the constructor. I'll make the change.

src/core/src/UsdNotificationHandler.cpp
```diff
@@ -180,203 +220,271 @@ void processCesiumTilesetChanged(const pxr::SdfPath& tilesetPath, const std::vec
...
+    if (updateBindings) {
+        updateImageryBindings(context, imageryPath);
     }
 
     if (updateImageryLayerAlpha) {
-        const auto imageryLayerIndex = tileset->findImageryLayerIndex(imageryPath);
-        if (imageryLayerIndex.has_value()) {
-            tileset->updateImageryLayerAlpha(imageryLayerIndex.value());
-        }
+        updateImageryBindingsAlpha(context, imageryPath);
```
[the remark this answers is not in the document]
lilleyse: No, because `updateImageryBindingsAlpha` doesn't refresh tilesets, it just calls `pTileset->updateImageryLayerAlpha(i)` which updates materials.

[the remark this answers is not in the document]
corybarr: Ah, good. I think I misread where CTRL-click jumped.

src/core/src/UsdUtil.cpp
```diff
@@ -1,63 +1,100 @@
...
 
-pxr::UsdStageRefPtr getUsdStage() {
-    return Context::instance().getStage();
-}
+namespace {
 
-long getUsdStageId() {
-    return Context::instance().getStageId();
-}
+class ScopedEdit {
```
corybarr, line 52: I wonder if class should live in the header file. It seems generally useful, so seeing it if perusing the headers sounds useful. Outside of some wrappers in FabricUtil.cpp, we don't declare other classes in .cpp files.

lilleyse: Yeah it could be useful outside of here. I moved it to its own file.

exts/cesium.usd.plugins/cesium/usd/plugins/CesiumUsdSchemas/__init__.pyi
corybarr: Yes, but in the past only partially successfully. I would get errors, but after some fixing with @weegeekps the errors didn't block compilation. Maybe the errors prevented some generated changes from happening. The error was consistent, but as of today I'm not able to reproduce them. Maybe there was a patch. I'll keep an eye on it next time I edit the schemas.

corybarr:
There's an issue in this branch that doesn't reproduce in `main`. If you de-parent a `CesiumPolygonImageryPrim` from a `CesiumTilesetPrim` by dragging it in the Stage Window, then when you re-parent the `CesiumPolygonImageryPrim` its polygon won't clip. To reproduce:

1. Open the USDA file. You'll see two clipped polygons:

image

2. Create a `Scope`
3. Drag a `CesiumPolygonImageryPrim` to the scope :

image

4. Drag the `CesiumPolygonImageryPrim` back to the `CesiumTilesetPrim`
5. Refresh the tileset:
image

[clippingTest.twoPolygonLayers.cleanupBranch.usda.zip](https://github.com/CesiumGS/cesium-omniverse/files/14069063/clippingTest.twoPolygonLayers.cleanupBranch.usda.zip)

lilleyse:
@corybarr that was fixed in https://github.com/CesiumGS/cesium-omniverse/pull/626/commits/be1024cac2de62f0a5e770136eab40530804dc0a. We must have both independently discovered the issue around the same time.

corybarr:
> @corybarr that was fixed in [be1024c](https://github.com/CesiumGS/cesium-omniverse/commit/be1024cac2de62f0a5e770136eab40530804dc0a). We must have both independently discovered the issue around the same time.

Already fixed. Awesome.

lilleyse:
I disabled building tests on CI - I was seeing linker errors that I couldn't reproduce locally and don't really have the energy to figure out. The test infrastructure needs to be rethought in order to be CI compatible.

```
[100%] Linking CXX shared module ../../lib/CesiumOmniverseCppTestsPythonBindings.cpython-310-x86_64-linux-gnu.so
lto-wrapper: warning: using serial compilation of 2 LTRANS jobs
/usr/bin/ld: ../../lib/libCesiumOmniverseCore.a(UsdTokens.cpp.o): in function `carb::cpp::atomic<carb::detail::CachedInterface<omni::fabric::IToken, (char const*)0>::RequestState>::wait(carb::detail::CachedInterface<omni::fabric::IToken, (char const*)0>::RequestState, std::memory_order) const':
UsdTokens.cpp:(.text._ZNK4carb3cpp6atomicINS_6detail15CachedInterfaceIN4omni6fabric6ITokenELPKc0EE12RequestStateEE4waitESA_St12memory_order[_ZNK4carb3cpp6atomicINS_6detail15CachedInterfaceIN4omni6fabric6ITokenELPKc0EE12RequestStateEE4waitESA_St12memory_order]+0xbd): undefined reference to `carb::thread::detail::ParkingLot::WaitBucket::bucket(void const*)::waitBuckets'
/usr/bin/ld: ../../lib/libCesiumOmniverseCore.a(UsdTokens.cpp.o): relocation R_X86_64_PC32 against undefined hidden symbol `_ZZN4carb6thread6detail10ParkingLot10WaitBucket6bucketEPKvE11waitBuckets' can not be used when making a shared object
/usr/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status
gmake[2]: *** [tests/bindings/CMakeFiles/CesiumOmniverseCppTestsPythonBindings.dir/build.make:159: lib/CesiumOmniverseCppTestsPythonBindings.cpython-310-x86_64-linux-gnu.so] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:678: tests/bindings/CMakeFiles/CesiumOmniverseCppTestsPythonBindings.dir/all] Error 2
gmake: *** [Makefile:166: all] Error 2
```

lilleyse:
@corybarr I don't think I can get to the sign out issue in this PR, though I would like to address it before the release.

I'm marking the PR as non-draft now. I think I've tested as much as I can and it would be good to get this finally merged.

corybarr (approved):
This codebase feels considerable more manageable now.

PR #634, opened 2024-01-25
asset accessor that uses libcurl directly

timoore:
This PR replaces the HttpAccessAccessor with the accessor from vsgCs, called UrlAssetAccessor. UrlAssetAccessor caches the CURL handle between requests, allowing libcurl to keep connections open to remote servers. This can be a big performance gain, especially for https connections. Apparently this was always problematic in the cpr library used by HttpAccessAccessor.

There was also some cleanup in the rest of the code to refer to CesiumAsync:IAssetAccessor instead of a concrete accessor type where possible.


weegeekps (commented):
Looks good, but lacks support for proxies, which I suspect will be an issue for academic and government customers.

There is a great overview on [everything.curl.dev](https://everything.curl.dev/transfers/conn/proxies) regarding supporting proxies. Curl actually does most of the heavy lifting here, but the hard part is actually gathering the correct settings for the proxy.

I want to add, I'm not entirely sure this is required. I believe that `libcpr` does give us proxy detection but I can't fully remember and if it doesn't then nothing is broken by us swapping to this.

weegeekps:
So it turns out CPR doesn't autoresolve proxies like I thought it did, so my comments about the proxy support are moot. Probably needs to be fixed one day.

src/core/src/UrlAssetAccessor.cpp
```diff
@@ -0,0 +1,288 @@
...
+    // XXX Do we need to worry about the thread safety problems with this?
+    curl_global_init(CURL_GLOBAL_ALL);
+}
+
+UrlAssetAccessor::~UrlAssetAccessor() {
+    curl_global_cleanup();
+}
+
+curl_slist*
+UrlAssetAccessor::setCommonOptions(CURL* curl, const std::string& url, const CesiumAsync::HttpHeaders& headers) {
```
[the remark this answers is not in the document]
timoore: > Below are a couple other easy opts that may be useful for performance. Admittedly two of these are taken from [the Performance page on everything.curl.dev](https://everything.curl.dev/libcurl/performance) but in my experience these are the most impactful performance tweaks.
> 
> * [CURLOPT_BUFFERSIZE](https://curl.se/libcurl/c/CURLOPT_BUFFERSIZE.html) controls the _preferred_ size of the receive buffer. The default (usually CURLOPT_MAX_WRITE_SIZE which is usually 16 KiB) is good enough for 90% of use cases, but for larger data like ours it is probably helpful to have it larger. My gut says 512 KiB or 1 MiB, but we may need to do a little testing here. Beware, set it once per handle in the pool, and never set it again, especially if that handle is currently being used for a connection. Bad things will happen.

Do you know if that's the same as CPR_RESERVE_SIZE? That's set to 3Mb. I will set that.
I'll add the other options.

src/core/src/UrlAssetAccessor.cpp
```diff
@@ -0,0 +1,288 @@
...
+    const CesiumAsync::AsyncSystem& asyncSystem,
+    const std::string& verb,
+    const std::string& url,
+    const std::vector<CesiumAsync::IAssetAccessor::THeader>& headers,
+    const gsl::span<const std::byte>& contentPayload) {
+    return asyncSystem.createFuture<std::shared_ptr<CesiumAsync::IAssetRequest>>([&](const auto& promise) {
+        auto request = std::make_shared<UrlAssetRequest>(verb, url, headers);
+        auto payloadCopy = std::make_shared<std::vector<std::byte>>(contentPayload.begin(), contentPayload.end());
+        asyncSystem.runInWorkerThread([promise, request, payloadCopy, this]() {
+            VSGCS_ZONESCOPEDN("UrlAssetAccessor::request inner");
```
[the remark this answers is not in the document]
timoore: Yes, #defined out, but still cruft. These were good tracing points in vsgCs; I'll see about replacing them with cesium-native tracing calls.

src/core/src/UrlAssetAccessor.cpp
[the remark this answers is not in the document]
weegeekps: I believe `CPR_RESERVE_SIZE` is specific to  CPR. That said, I haven't checked but I wouldn't be surprised if under the covers CPR just passes the value to libcurl using this opt.

lilleyse:
I'm seeing SSL errors on Windows:

> 2024-01-26 18:20:32 [24,003ms] [Error] [cesium.omniverse.plugin] [2024-01-26 13:20:32.916] [cesium-omniverse] [error] [RasterOverlayCollection.cpp:127] Error while creating tile provider: curl: SSL peer certificate or SSH remote key was not OK
>
> 2024-01-26 18:20:32 [24,005ms] [Error] [cesium.omniverse.plugin] [2024-01-26 13:20:32.918] [cesium-omniverse] [error] Error while creating tile provider: curl: SSL peer certificate or SSH remote key was not OK       
>
> 2024-01-26 18:20:32  [Error] [cesium.omniverse.plugin] [2024-01-26 13:20:32.922] [cesium-omniverse] [error] [TilesetContentManager.cpp:825] An unexpected error occurred when loading tile: curl: SSL peer certificate or SSH remote key was not OK

timoore:
> I'm seeing SSL errors on Windows:
> 
I've seen reports of this on Windows with vsgCs, but it has always worked for me. I think it's sensitive to the exact version of curl+ssl that is downloaded. I will check if cpr has some explicit dependency on SSL / TLS that isn't being satisfied now.

timoore:
Sean, can you try it on Windows with the latest commit? We thought that the local certificate was redundant, but maybe there are additional hopes to jump through to use OS provided certificates.

I would try it myself, but I'm in the airport and don't have any Windows environment installed.

lilleyse:
The latest commit fixes it.

It would still be good to know why it's needed for Windows. I was hoping we could get away without shipping certs.

timoore:
https://stackoverflow.com/questions/37551409/configure-curl-to-use-default-system-cert-store-on-windows looks promising.

weegeekps:
> Starting with curl 8.2.0 (July 19 2023) (issue), there is the option --ca-native

I wonder how that option translates from CLI to libcurl?

timoore:
Another try, this time with the CURLSSOPT_NATIVE_CA option. @lilleyse can you give the latest a quick try?

lilleyse:
@timoore looks like there's a merge conflict now that #626 is merged. I think the strategy would be to take the changes from `main` and re-do the smaller changes that you made in `Context`, etc.

Also, could you remove `HttpAssetAccessor` and anything that mentions `cacert.pem` including the `update_certs` script? We don't need to install the `requests` library any more either. That can be removed from the developer setup and `main.yml`.

r-veenstra:
I did some load time testing on this branch and observed some very positive results compared to current main, and also the version of main prior to this branch fork (to ensure the massive refactor didn't impact anything)

Process:
- USDA with a fixed camera position over Manhattan & Google 3D tiles
- Hacked in a load timer to the `Cesium Debugging` window so `Reload all Tilesets` would trigger a timer that incremented as the tiles load
- Verified each test was loading the exact same quantity of tiles (2382) and fetching data across the network (~100mb for each reload)

**Current main:** f89fa4e4a33a2e68d6842fe46adf70065b0d06a2
Test 1:	92.85
Test 2:	85.75
Test 3:	87.22
Test 4:	88.51
Test 5:	86.78
**Average:	88.222**

**This Branch:** e602ddd5215330204d13f211bec74648555d7b3a
Test 1:	52.79
Test 2:	51.25
Test 3:	52.5
Test 4:	51.07
Test 5:	51.7
**Average:	51.862**

**Main pre-libcurl:** d1b7ebcbdcd0020fca64eb21cc8a19ed3abe6a9d
Test 1:	89.67
Test 2:	86.9
Test 3:	88.7
Test 4:	86.89
Test 5:	86.52
**Average:	87.736**

Unless I've made a mistake, that's a 35% load time reduction for a complex Google 3D tiles scene.

src/core/src/Context.cpp
```diff
@@ -67,10 +66,6 @@ const std::filesystem::path& Context::getCesiumExtensionLocation() const {
     return _cesiumExtensionLocation;
 }
 
-const std::filesystem::path& Context::getCertificatePath() const {
-    return _certificatePath;
-}
```
lilleyse, line 72: Remove `getCertificatePath` from `Context.h`

src/core/include/cesium/omniverse/UrlAssetAccessor.h
```diff
@@ -0,0 +1,130 @@
...
+                entry.free = true;
+                return;
+            }
+        }
+        throw std::logic_error("releasing a curl handle that is not in the cache");
+    }
+};
+
+// Simple implementation of AssetAcessor that can make network and local requests
+class UrlAssetAccessor : public CesiumAsync::IAssetAccessor {
```
lilleyse, line 81: `UrlAssetAccessor`, `UrlAssetResponse`, `UrlAssetRequest` could all be marked `final`

src/core/include/cesium/omniverse/UrlAssetAccessor.h
```diff
@@ -0,0 +1,130 @@
...
+    CesiumAsync::Future<std::shared_ptr<CesiumAsync::IAssetRequest>> request(
+        const CesiumAsync::AsyncSystem& asyncSystem,
+        const std::string& verb,
+        const std::string& url,
+        const std::vector<CesiumAsync::IAssetAccessor::THeader>& headers,
+        const gsl::span<const std::byte>& contentPayload) override;
+
+    void tick() noexcept override;
+    CurlCache curlCache;
+    std::string userAgent;
```
lilleyse, line 100: Can `userAgent` be private?

src/core/include/cesium/omniverse/UrlAssetAccessor.h
```diff
@@ -0,0 +1,130 @@
...
+// A cache that permits reuse of CURL handles. This is extremely important for performance
+// because libcurl will keep existing connections open if a curl handle is not destroyed
+// ("cleaned up").
+
+struct CurlCache {
+    struct CacheEntry {
+        CacheEntry()
+            : curl(nullptr)
+            , free(false) {}
+        CacheEntry(CURL* in_curl, bool in_free)
```
lilleyse, line 48: In cesium-native I see the convention `free_` a lot

src/core/include/cesium/omniverse/UrlAssetAccessor.h
```diff
@@ -0,0 +1,130 @@
...
+            : curl(in_curl)
+            , free(in_free) {}
+        CURL* curl;
+        bool free;
+    };
+    std::mutex cacheMutex;
+    std::vector<CacheEntry> cache;
```
lilleyse, line 55: Do the `CURL` objects need to be explicitly freed somewhere? Like when CurlCache is destroyed?

src/core/src/UrlAssetAccessor.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+#include <CesiumAsync/IAssetResponse.h>
+#include <CesiumUtility/Tracing.h>
+#include <omni/kit/IApp.h>
+
+#include <algorithm>
+#include <cstring>
+
+namespace cesium::omniverse {
+const auto CURL_BUFFERSIZE = 3145728L; // 3 MiB
```
lilleyse, line 14: Do you know when the buffer gets freed? Is it when `curl_easy_reset` is called, or when the `CURL` object is destroyed? Hopefully not the former.

src/core/src/UrlAssetAccessor.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+        _headers.insert({"X-Cesium-Client", "Cesium for Omniverse"});
+        _headers.insert(
+            {"X-Cesium-Client-Version", fmt::format("v{} {}", CESIUM_OMNI_VERSION, CESIUM_OMNI_GIT_HASH_ABBREVIATED)});
+        _headers.insert({"X-Cesium-Client-Engine", fmt::format("Kit SDK {}", buildInfo.kitVersion)});
+        _headers.insert({"X-Cesium-Client-OS", platformInfo.platform});
+    }
+
+    const std::string& method() const override {
+        return this->_method;
```
lilleyse, line 70: Style nitpick: convention in this repo is to omit `this->`. Though looks like native is different.

src/core/src/UrlAssetAccessor.cpp
```diff
@@ -0,0 +1,268 @@
...
+CesiumAsync::Future<std::shared_ptr<CesiumAsync::IAssetRequest>> UrlAssetAccessor::get(
+    const CesiumAsync::AsyncSystem& asyncSystem,
+    const std::string& url,
+    const std::vector<CesiumAsync::IAssetAccessor::THeader>& headers) {
+    return asyncSystem.createFuture<std::shared_ptr<CesiumAsync::IAssetRequest>>([&](const auto& promise) {
+        std::shared_ptr<UrlAssetRequest> request = std::make_shared<UrlAssetRequest>("GET", url, headers);
+        asyncSystem.runInWorkerThread([promise, request, this]() {
+            CESIUM_TRACE("UrlAssetAccessor::get");
+            CurlHandle curl(this);
+            curl_slist* list = setCommonOptions(curl(), request->url(), request->headers());
```
lilleyse, line 193: What's the overhead of calling `setCommonOptions`? Many of the options won't change per-request. Could some of them be set when acquiring a new handle?

If there's minimal overhead then no need to change it.

src/core/src/UrlAssetAccessor.cpp
lilleyse: Also I think this could be shrunk to 0.5 MiB given the avg tile sizes in CWT and Google.

src/core/include/cesium/omniverse/UrlAssetAccessor.h
timoore: I prefer that too and will make the change.

src/core/include/cesium/omniverse/UrlAssetAccessor.h
timoore: They don't need to be freed for the duration of the program. It's complicated to figure out the multi-threading implications of when a CURL object is in use or may be freed, and this seems to be a weakness in the curl design. I thought it was best to blow off freeing the CURL objects and just let them be deleted when the program stops.

src/core/src/UrlAssetAccessor.cpp
timoore: I believe curl_easy_reset() doesn't do much except to prepare for another request. For example, it doesn't tear down an existing connection. The buffer freeing should probably be verified, either by examining the code or with a debugger.

src/core/src/UrlAssetAccessor.cpp
timoore: I prefer to omit `this->` too. I probably copied the code directly from cesium-unreal.

src/core/src/UrlAssetAccessor.cpp
timoore: I believe that it is harmless to set the options each time and that it's not too obvious from the docs what options persist after the curl_easy_reset() call. Merits further examination.

lilleyse:
@timoore could you update `CHANGES.md` summarizing the performance improvements that Ryan noted: https://github.com/CesiumGS/cesium-omniverse/pull/634#issuecomment-1916288094

lilleyse (approved):
Looks good. Just update `CHANGES.md`, `ThirdPart.extra.json`, and copyrights, and then you can go ahead and merge.

PR #645, opened 2024-01-29
Raster overlays as `rel` instead of child prims

### LINKED ISSUE #585: Use `rel` for imagery layers
Currently imagery layers are child prims of a tileset, but it would make more sense for them to be components of a tileset like in Cesium for Unreal and Unity. We might be able to use a `multipleApply` schema for this. This would be a breaking change, so we should try to do it in the same release as the imagery related breaking changes for tileset clipping.

Some questions:

* Can the order of imagery layers be rearranged in the UI when using `multipleApply` schema?

CC @weegeekps @corybarr

### LINKED ISSUE #586: Make imagery-layer order editable in the UI
Currently, it's not possible to order the imagery layers in the UI. In the example below, I want to move the `CesiumPolygonImageryLayer` to be after the `CesiumIonImageryLayer`. Currently, customers would need to edit the USDA file directly.

image

lilleyse:
Fixes #585
Fixes #586 

Raster overlay are now attached to tilesets through the `cesium:rasterOverlayBinding` property. Previously, raster overlays needed to be children of a tileset. This has a few advantages:

* Multiple tilesets can share the same raster overlay
* It's easier to re-order raster overlays
* This is more the USD way to do things

https://github.com/CesiumGS/cesium-omniverse/assets/915398/46af8f48-0565-476b-9df3-322854bae7eb

This video shows off a few different things:

* Set the georeference to Washington DC
* Create CWT + Bing Imagery
* Move Bing Imagery to a different folder and confirm that the `rel` points to the new location
* Create a second raster overlay for Washington DC imagery and add it to CWT
* Check that changing alpha still works
* Check that reordering the raster overlays works
* Add Google 3D Tiles
* Add Washington DC imagery to Google 3D Tiles. Now both tilesets are referencing this raster overlay.
* Delete raster overlay bindings

One thing I forgot to do in the video but checked later was to delete the raster overlay prim and later re-add it at the same path. The tilesets get updated correctly when it is removed and re-added.


src/core/src/OmniTileset.cpp
```diff
@@ -485,47 +493,53 @@ void OmniTileset::reload() {
...
 }
 
-uint64_t OmniTileset::getRasterOverlayLayerCount() const {
-    return _rasterOverlayPaths.size();
-}
+void OmniTileset::updateRasterOverlayAlpha(const pxr::SdfPath& rasterOverlayPath) {
+    const auto rasterOverlayPaths = getRasterOverlayPaths();
+    const auto rasterOverlayIndex = CppUtil::indexOf(rasterOverlayPaths, rasterOverlayPath);
+
+    if (rasterOverlayIndex == rasterOverlayPaths.size()) {
```
corybarr, line 529: Any case where `>=` would be true but not `==`?

src/core/src/OmniTileset.cpp
```diff
@@ -485,47 +493,53 @@ void OmniTileset::reload() {
...
-void OmniTileset::updateRasterOverlayLayerAlpha(uint64_t rasterOverlayLayerIndex) {
-    assert(rasterOverlayLayerIndex < _rasterOverlayPaths.size());
+    const auto pRasterOverlay = _pContext->getAssetRegistry().getRasterOverlay(rasterOverlayPath);
+
+    if (!pRasterOverlay) {
+        return;
+    }
 
-    const auto alpha = getRasterOverlayLayerAlpha(rasterOverlayLayerIndex);
+    const auto alpha = glm::clamp(pRasterOverlay->getAlpha(), 0.0, 1.0);
```
corybarr, line 539: This is also clamped via the UI in an incoming PR, but this is a nice guard.

corybarr (approved):

src/core/src/OmniTileset.cpp
lilleyse: It shouldn't ever happen. Kind of like how `std::find` returns the end iterator if not found, `indexOf` returns the size of the vector.

PR #648, opened 2024-01-30
Use rel for cartographic polygon

### LINKED ISSUE #632: Cartographic Polygon editing crash - fabric only
1. Start USD Composer (I used 2023.2.1) - build from commit d1b7ebcbdcd0020fca64eb21cc8a19ed3abe6a9d
2. Enable fabric
3. Add a Cartographic Polygon via the quick add menu
4. Select it
5. Click Edit Control Vertices
6. Click on the default ground plane to place a first point
7. OV should briefly hang then crash

Note: Does not crash when fabric disabled

cc @corybarr

### LINKED ISSUE #639: Use `rel` for CesiumCartographicPolygon
Based on discussion in https://github.com/CesiumGS/cesium-omniverse/issues/632 there a few disadvantages of `CesiumCartographicPolygon` being a subclass of `BasisCurves`. We should consider using `rel` instead.

corybarr:
Resolves #639 
Fixes #632 

Based on the plan to add the `CesiumCartographicPolygonPrim` back into the codebase once the issue with `BasisCurves` inheritance is fixed, these are the minimal set of code changes to achieve removing the prim. To the end user, the "cartographic polygon" concept and terminology still exists. The term is still used in the UI. The only difference to the end user once the prim is added back should be the type of the prim in the Stage Window.

This means that `OmniCartographicPolygon` still exists, even though there is no corresponding USD prim.

Test file: [noCartographicPolygon.clippingTest.TwoOverlays.usda.zip](https://github.com/CesiumGS/cesium-omniverse/files/14093903/noCartographicPolygon.clippingTest.TwoOverlays.usda.zip)


corybarr:
> I've been testing and appears to be working as expected. All I could probably suggest is a warning to the log if a user links a basis curve without a globe anchor (something I've already done by accident).

I can see the use of that. Let's create an issue during standup if we want to. I think our `rel` workaround is good as workarounds go, but ultimately we should inherit from a `BasisCurves` so we can define the `GlobeAnchor` requirement in the schema. So any warnings would be temporary with an issue that keeps us from forgetting to remove them in the future.

lilleyse (approved):
Looks great! Nothing from me.

PR #649, opened 2024-01-30
Raster overlay UI/UX enhancements

### LINKED ISSUE #618: Enhance UI/UX for raster overlays
This entails the following:

- Make the UI have feature parity with Cesium for Unreal
- Add custom UI widgets for each raster-overlay type
- Make sure the dialog window is filtering the list of prims properly by type when selecting a `rel`
- Appropriate interfaces for the values where possible
  - e.g., a slider bounded to [0,1] for alpha values

### LINKED ISSUE #631: Cartographic Polygon and Polygon Imagery layer default values
I suggest we implement the following default values to make using cartographic polygons and polygon overlays a little easier

Imagery Overlay
- [x] `Overlay Render Method - clipping` - same as Unreal, also less confusing as it won't immediately make the tileset black

Cartographic Polygons - defaults when creating by any method
- [x] `point3f[] points = [(-1000, 0, -1000), (-1000, 0, 1000), (1000, 0, 1000), (1000, 0, -1000)]` - Helpful to have some pre-existing points for the user to grab and edit, vs trying to draw from scratch.  Unreal does this. (especially as raycasting to tilesets is not currently available)
- [x] `uniform token type = "linear"` - Helpful to avoid accidental use of curves which aren't supported
- [x] `uniform token wrap = "periodic"` - Ensures the polygon is closed

cc @corybarr

### LINKED ISSUE #633: Invert option for polygon raster overlays
Add support for inverting the polygon raster overlay as per Cesium for Unreal.  Make clipping exclude outside of the polygon instead of inside.

image

image

cc @corybarr

corybarr:
- Adds custom UI widgets for `CesiumPolygonRasterOverlayPrim `and `CesiumIonRasterOverlayPrim`
- Adds "Invert Selection" to `CesiumPolygonRasterOverlayPrim`
- Makes `CesiumPolygonRasterOverlayPrim `and `CesiumIonRasterOverlayPrim` "alpha" attribute a slider
- Makes `BasisCurves` created via Quick-Add menu have appropriate settings for `type`, `wrap`, and `points` attributes.
- Sets ordering of UI groups to match Unreal (polygons->invert->rendering->credits)
- Constrains cartographic polygon bindings to only be `BasisCurves`

image

Resolves #633 
Resolves #618
Resolves #631


src/core/src/OmniPolygonRasterOverlay.cpp
```diff
@@ -85,7 +85,20 @@ void OmniPolygonRasterOverlay::reload() {
...
+    const auto prim = usdStage->GetPrimAtPath(path);
+    if (prim.IsValid()) {
+        const auto cesiumPolygonRasterOverlay = UsdUtil::getCesiumPolygonRasterOverlay(_pContext->getUsdStage(), _path);
+        auto attr = cesiumPolygonRasterOverlay.GetInvertSelectionAttr();
+        if (attr.IsValid()) {
+            bool value;
+            attr.Get(&value);
+            invertSelection = value;
+        }
+    }
```
lilleyse, line 100: Could you add a member function for this? See `OmniRasterOverlay::getShowCreditsOnScreen` for an example. Some things like checking prim and attr validity may not be needed.

exts/cesium.omniverse/cesium/omniverse/usdUtils/usdUtils.py
```diff
@@ -103,7 +103,13 @@ def add_cartographic_polygon() -> str:
     cartographic_polygon_path = Sdf.Path("/CesiumCartographicPolygons").AppendPath(name).pathString
     cartographic_polygon_path = omni.usd.get_stage_next_free_path(stage, cartographic_polygon_path, False)
 
-    UsdGeom.BasisCurves.Define(stage, cartographic_polygon_path)
+    basis_curves_prim = UsdGeom.BasisCurves.Define(stage, cartographic_polygon_path).GetPrim()
+    basis_curves_prim.GetAttribute("type").Set("linear")
+    basis_curves_prim.GetAttribute("wrap").Set("periodic")
+    basis_curves_prim.GetAttribute("points").Set(
+        [(-1000, 0, -1000), (-1000, 0, 1000), (1000, 0, 1000), (1000, 0, -1000)]
+    )
```
lilleyse, line 111: Could you use the `UsdGeom.BasisCurves` functions instead of `GetAttribute`? There should be `GetTypeAttr`, `GetWrapAttr`, `GetPointsAttr`.

exts/cesium.omniverse/cesium/omniverse/ui/attributes/ion_raster_overlay_attributes_widget.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def _customize_props_layout(self, props):
+        frame = CustomLayoutFrame(hide_extra=True)
+
+        with frame:
+
+            def update_range(stage, prim_paths):
+                for path in prim_paths:
+                    prim = stage.GetPrimAtPath(path)
+                    attr = prim.GetAttribute("cesium:alpha") if prim else None
```
lilleyse, line 30: Avoid using `GetAttribute` here too.

```python
rasterOverlay = CesiumRasterOverlay.Get(stage, path)
attr = rasterOverlay.GetAlphaAttr()
```

exts/cesium.omniverse/cesium/omniverse/ui/attributes/ion_raster_overlay_attributes_widget.py
```diff
@@ -0,0 +1,100 @@
...
+                        value_widget = UsdPropertiesWidgetBuilder.create_drag_or_slider(
+                            ui.FloatDrag, ui.FloatSlider, **widget_kwargs
+                        )
+                        mixed_overlay = UsdPropertiesWidgetBuilder.create_mixed_text_overlay()
+                    UsdPropertiesWidgetBuilder.create_control_state(
+                        value_widget=value_widget, mixed_overlay=mixed_overlay, **widget_kwargs
+                    )
+                    model.add_value_changed_fn(lambda m, s=stage, p=prim_paths: update_range(s, p))
+
+                    return model
```
lilleyse, line 77: Is it possible to make this more generic and put in a helper file so it's not duplicated across files? E.g. I imagine the min/max values would be provided as options instead of hardcoded.

exts/cesium.usd.plugins/schemas/cesium_schemas.usda
```diff
@@ -500,6 +500,16 @@ class CesiumPolygonRasterOverlayPrim "CesiumPolygonRasterOverlayPrim" (
         displayName = "Cartographic Polygon Binding"
         doc = "Specifies which Cartographic Polygons to use in the raster overlay"
     )
+
+    bool cesium:invertSelection = false (
```
lilleyse, line 504: Any new USD fields need to be accounted for in `UsdNotificationHandler`. In this case, should be a simple change in `processCesiumPolygonRasterOverlayChanged`.

exts/cesium.omniverse/cesium/omniverse/usdUtils/usdUtils.py
corybarr: I actually had those, but `GetTypeAttr` threw an error that it's not a member of `USDGeom.BasisCurves` so I resorted to the string. Let me double check that now that the code is more stabilized.

lilleyse:
This doesn't need to be addressed now...

If I set a polygon overlay's render method to `overlay` it appears red. Is this something we have control over, or are we just rendering the pixel color exactly as it comes from native?

image

corybarr:
> This doesn't need to be addressed now...
> 
> If I set a polygon overlay's render method to `overlay` it appears red. Is this something we have control over, or are we just rendering the pixel color exactly as it comes from native?
> 
> image

We don't have control over it right now, but it's not a major lift. I created a ticket for it. #652

exts/cesium.omniverse/cesium/omniverse/usdUtils/usdUtils.py
corybarr: I probably tried to use `SetTypeAttr`, which doesn't exist. At any rate, is this everyone's current state of IntelliSense in the Python code? 

image

lilleyse: Yes, I think it's a problem with Omniverse's Python bindings

r-veenstra:
- Invert selection appears to obey the initial setting on load, but when changing the bool and refreshing the tileset it does not update accordingly.  If I save the USDA with the updated bool and reload the stage, it works as expected.
- Default values for the cartographic polygons appear to be good
- Changing Invert Selection should probably trigger a tileset refresh (similar to the refresh that occurs when adding or removing polygons from the raster overlay).  Though that could be frustrating unless https://github.com/CesiumGS/cesium-omniverse/issues/650 is fixed first

You should be able to test the invert issue with the following zip

[TilesetClipping.zip](https://github.com/CesiumGS/cesium-omniverse/files/14107357/TilesetClipping.zip)

r-veenstra:
Thanks, invert issues appear resolved now!

r-veenstra:
I added `basis_curves.GetCurveVertexCountsAttr().Set([4])` to the creation of basis curves, as without this property the curves never render, and the control points never show when trying to edit them

exts/cesium.omniverse/cesium/omniverse/ui/attributes/ion_raster_overlay_attributes_widget.py
corybarr: OK. Pushed. You can now call `CustomLayoutProperty("cesium:alpha", build_fn=build_slider(0, 1))` with whichever values you want. 

There might be a more Pythonic way to do this (with `partial`?) but this feels OK. I'm not yet fluently Pythonic, and some of my Python I've been told is like transliterated game-engine C++.  This is literally transliterated game-engine C# code: `[Range(0,1)]`.

src/core/src/OmniPolygonRasterOverlay.cpp
```diff
@@ -96,8 +102,9 @@ void OmniPolygonRasterOverlay::reload() {
         _pContext->getLogger()->error(error.message);
     };
 
+    auto debugInvert = getInvertSelection();
     _pPolygonRasterOverlay = new CesiumRasterOverlays::RasterizedPolygonsOverlay(
-        rasterOverlayName, polygons, invertSelection, *pEllipsoid, projection, options);
+        rasterOverlayName, polygons, debugInvert, *pEllipsoid, projection, options);
```
lilleyse, line 107: Why is it called `debugInvert`?

exts/cesium.omniverse/cesium/omniverse/ui/attributes/custom_attribute_widgets.py
```diff
@@ -0,0 +1,70 @@
...
+)
+
+
+def _update_range(stage, prim_paths):
+    for path in prim_paths:
+        rasterOverlay = CesiumRasterOverlay.Get(stage, path)
+        attr = rasterOverlay.GetAlphaAttr()
+        current_value = attr.Get()
+        new_value = max(0, min(current_value, 1.0))
+        attr.Set(new_value)
```
lilleyse, line 15: Is it possible to not hardcode `CesiumRasterOverlay`? I imagine we'll want to use this widget for other prim types.

Also it's not clear to me how the min/max options that are passed to `build_slider` are getting set. Seems like 0 and 1 are still hardcoded in a couple places.

src/core/src/OmniPolygonRasterOverlay.cpp
corybarr: Thanks. I forgot to change that back. I was debugging inversion not working, fixed it, wrapped up for the day, and forgot this line.

exts/cesium.omniverse/cesium/omniverse/ui/attributes/custom_attribute_widgets.py
corybarr: Ah, I'd left 0 and 1 in `widget_kwargs`. Gone now. 

I was able to remove the dependency on `CesiumRasterOverlay`. 

Note: There were issues operating on generic prims. For some reason, only class functions to access attributes would work (e.g., `GetAlphaAttr`) in event handlers. Fortunately, we don't need `update_range` anyways. It's cruft from an example where bounds and its associated attributes needed to be updated. We might possible run into this in the future if we want to do something like update bounds on generic prims.

lilleyse (approved):

PR #655, opened 2024-01-31
Only print warnings for properties that are in use

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/588

* https://github.com/CesiumGS/cesium-omniverse/commit/d02944fd13be9d8633ae97733770c047b343c91a - cleanup commit so that `oneTimeWarning` can take multiple parameters
* https://github.com/CesiumGS/cesium-omniverse/commit/c5c68dc412801532b789b16efd9f4f7991a050a6 - use `oneTimeWarning` for "Could not find property" and "incompatible type" warnings

### DIFF
--- a/src/core/include/cesium/omniverse/CppUtil.h
+++ b/src/core/include/cesium/omniverse/CppUtil.h
@@ -4,6 +4,7 @@
 #include <array>
 #include <cstdint>
 #include <functional>
+#include <map>
 #include <optional>
 #include <set>
 #include <unordered_set>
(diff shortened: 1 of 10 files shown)


lilleyse:
I'm working on a more complete fix for https://github.com/CesiumGS/cesium-omniverse/issues/588, marked as draft in the meantime.

lilleyse:
Looking into a performance degradation with the new change

PR #656, opened 2024-01-31
Update CHANGES.md

lilleyse:
Rendered preview: https://github.com/CesiumGS/cesium-omniverse/blob/update-changes/CHANGES.md

@corybarr @r-veenstra let me know if anything is missing here


corybarr:
I can't think of anything omitted.

If someone has a local tileset like I do in a hobby project, they might be wondering what type of prim to use. `CesiumIonRasterOverlayPrim` should work, but it's got irrelevant properties and an unintuitive name. Do we need a different prim for these tilesets?

corybarr (approved):
I can't think of anything left out.

PR #657, opened 2024-01-31
Increase width of basis curves

r-veenstra:
This PR increases the width of basis curves created by the Cesium Cartographic Polygon Quick Add option to 0.5m.  This size makes the curve easier to manipulate for **typical** use cases such as removing city blocks, buildings, etc.

image

Both the width and the default 10m x 10m dimension have been configured to be consistent based on the stage units being used (previously a 1000m x 1000m curve would be created if a user was using stage units = 1 meter.


corybarr (approved):
Tested and looks good.

PR #658, opened 2024-02-01
Bypass loading of clipped tiles

corybarr:
Add the following to functionality to polygon raster overlays:

image

(It seems tooltips don't wrap if they go off the screen.)


lilleyse:
@corybarr given Kevin's comment in https://github.com/CesiumGS/cesium-native/issues/803#issuecomment-1925986441, it may be possible to test this with another dataset like Melbourne Photogrammetry (in the asset depot).

corybarr:
> @corybarr given Kevin's comment in [CesiumGS/cesium-native#803 (comment)](https://github.com/CesiumGS/cesium-native/issues/803#issuecomment-1925986441), it may be possible to test this with another dataset like Melbourne Photogrammetry (in the asset depot).

@lilleyse Good call. It works well and feels very snappy loading just this section:

image

Logs confirm it works.

src/core/src/OmniPolygonRasterOverlay.cpp
```diff
@@ -16,7 +16,8 @@
 namespace cesium::omniverse {
 
 OmniPolygonRasterOverlay::OmniPolygonRasterOverlay(Context* pContext, const pxr::SdfPath& path)
-    : OmniRasterOverlay(pContext, path) {}
+    : OmniRasterOverlay(pContext, path)
+    , _pExcluder(nullptr) {}
```
lilleyse, line 20: `shared_ptr` are default constructed as nullptr, so `_pExcluder(nullptr)` isn't necessary.

It would be a different story if `_pExcluder` were a raw pointer.

src/core/src/OmniPolygonRasterOverlay.cpp
```diff
@@ -104,6 +116,14 @@ void OmniPolygonRasterOverlay::reload() {
...
     _pPolygonRasterOverlay = new CesiumRasterOverlays::RasterizedPolygonsOverlay(
         rasterOverlayName, polygons, getInvertSelection(), *pEllipsoid, projection, options);
+
+    auto cesiumPolygonRasterOverlay = UsdUtil::getCesiumPolygonRasterOverlay(_pContext->getUsdStage(), _path);
+    bool excludeSelectedTiles;
+    cesiumPolygonRasterOverlay.GetExcludeSelectedTilesAttr().Get(&excludeSelectedTiles);
+    if (excludeSelectedTiles) {
+        auto pPolygons = _pPolygonRasterOverlay.get();
+        _pExcluder = std::make_shared<Cesium3DTilesSelection::RasterizedPolygonsTileExcluder>(pPolygons);
+    }
```
lilleyse, line 126: Could be condensed to something like:

```c++
    if (getExcludeSelectedTiles()) {
        _pExcluder = std::make_shared<Cesium3DTilesSelection::RasterizedPolygonsTileExcluder>(_pPolygonRasterOverlay.get());
    }
```

src/core/src/OmniTileset.cpp
```diff
@@ -458,6 +458,7 @@ void OmniTileset::reload() {
     options.culledScreenSpaceError = getCulledScreenSpaceError();
     options.mainThreadLoadingTimeLimit = getMainThreadLoadingTimeLimit();
     options.showCreditsOnScreen = getShowCreditsOnScreen();
+    options.excluders = std::vector<std::shared_ptr<Cesium3DTilesSelection::ITileExcluder>>();
```
lilleyse, line 461: No need to manually initialize the vector. This line can be removed.

src/core/src/OmniTileset.cpp
```diff
@@ -504,6 +505,30 @@ void OmniTileset::reload() {
...
+        }
+
+        const auto exclude = pOmniPolygonRasterOverlay->getExcludeSelectedTiles();
+        if (exclude) {
+            const auto excluder = pOmniPolygonRasterOverlay->getExcluder();
+            if (excluder) {
+                _pTileset->getOptions().excluders.push_back(excluder);
+            }
+        }
+    }
```
lilleyse, line 531: This could be simplified to something like:

```c++
    const auto rasterOverlayPaths = getRasterOverlayPaths();

    for (const auto& rasterOverlayPath : rasterOverlayPaths) {
        const auto pPolygonRasterOverlay = _pContext->getAssetRegistry().getPolygonRasterOverlay(rasterOverlayPath);
        if (pPolygonRasterOverlay) {
            const auto pExcluder = pPolygonRasterOverlay->getExcluder();
            if (pExcluder) {
                options.excluders.push_back(pExcluder);
            }
        }
    }
```

and moved above so that it's not modifying options after the tileset has been constructed.

src/core/src/OmniTileset.cpp
```diff
@@ -504,6 +505,30 @@ void OmniTileset::reload() {
...
+
+        if (matchIterator == boundNativeRasterOverlays.end()) {
+            continue;
+        }
+
+        const auto exclude = pOmniPolygonRasterOverlay->getExcludeSelectedTiles();
+        if (exclude) {
+            const auto excluder = pOmniPolygonRasterOverlay->getExcluder();
+            if (excluder) {
+                _pTileset->getOptions().excluders.push_back(excluder);
```
lilleyse, line 528: Do excluders need to be manually removed in `destroyNativeTileset`? I needed to do that for raster overlays to avoid a crash. I haven't had a chance to debug yet and see if it's needed.

corybarr:
> > @corybarr given Kevin's comment in [CesiumGS/cesium-native#803 (comment)](https://github.com/CesiumGS/cesium-native/issues/803#issuecomment-1925986441), it may be possible to test this with another dataset like Melbourne Photogrammetry (in the asset depot).
> 
> @lilleyse Good call. It works well and feels very snappy loading just this section:
> 
> image
> 
> Logs confirm it works.

Clipping test for Melbourne: [MelbourneClippingTest.zip](https://github.com/CesiumGS/cesium-omniverse/files/14196972/MelbourneClippingTest.zip)

src/core/src/OmniTileset.cpp
corybarr: Interesting. In C#, you have to initialize their version of `std::vector`, but it's slightly different due to the `new`.

src/core/src/OmniTileset.cpp
lilleyse: Seems like it's working fine without.

src/core/src/OmniTileset.cpp
corybarr: Yeah, that's a lot simpler than querying the `RasterOverlayCollection`.

src/core/src/OmniTileset.cpp
corybarr: Yeah, I don't think it's needed. I saw that removal code in Unreal, but figured since they're a `std::vector<std::shared_ptr>` then they should destroy because in Omniverse the raster overlays don't talk to the tilesets.

src/core/src/OmniPolygonRasterOverlay.cpp
corybarr: Yes, the accessor is better. I added that before the accessor. It's probably good practice to set up every accessor for USD properties of prims when creating the Omni*.cpp classes.

lilleyse:
@corybarr I tested that https://github.com/CesiumGS/cesium-native/pull/808 fixes the clipping issues with P3DT.

corybarr:
> @corybarr I tested that [CesiumGS/cesium-native#808](https://github.com/CesiumGS/cesium-native/pull/808) fixes the clipping issues with P3DT.

Awesome. Will update and merge after that fix merged.

lilleyse:
@corybarr could you update `CHANGES.md`?

corybarr:
> @corybarr could you update `CHANGES.md`?

@lilleyse pushed

lilleyse:
@corybarr could you update cesium-native in this branch now that https://github.com/CesiumGS/cesium-native/pull/808 is merged?

corybarr:
> @corybarr could you update cesium-native in this branch now that [CesiumGS/cesium-native#808](https://github.com/CesiumGS/cesium-native/pull/808) is merged?

cesium-native is updated. The fix is confirmed working: 
image

lilleyse (approved):

Issue #650, opened 2024-01-30
Refresh tileset should not re-download tile data


r-veenstra:
Unsure if this is by design, however it is different to Cesium for Unreal at least.

When I click "Refresh Tileset" in a tilesets properties, this appears to trigger a full re-download of the tiles.  In some cases (particularly Google Photorealistic 3D tiles) this could be ~100mb of data over my network for a simple view.

![image](https://github.com/CesiumGS/cesium-omniverse/assets/123468416/01cfd0d0-173c-42f9-b8cb-9ce58f36e873)

Cesium for Unreal for comparison does not appear to be downloading data over my network when refreshing.

Given Refresh Tileset is often used when working with custom materials, `PolygonRasterOverlays`, and other workflows it would be ideal if it reloaded the cached data from disk.  This should improve refresh time substantially.

I have also verified my tileset cache is sufficiently large (bumped to 53687091200)

Data downloaded after a single click on `Refresh Tileset`
![image](https://github.com/CesiumGS/cesium-omniverse/assets/123468416/8a752d66-6b86-4811-ad3b-6ccb95090878)

Data downloaded after a second click on `Refresh Tileset`
![image](https://github.com/CesiumGS/cesium-omniverse/assets/123468416/af5099b9-0f51-4e8f-a5ca-5a25610dc650)

Unreal does not change here, and it also is significantly faster when a refresh is performed.

Local caching does appear to be working in OV when simply jumping the camera location from one place to another, and back again, as I don't get any data download then.  Just when clicking Refresh.

lilleyse:
@timoore could you take a look at this?

timoore:
Cesium-Unreal is using CesiumAsync::CachingAssetAccessor, which delegates to the accessor class making the http requests. In cesium-unreal, that eventually calls Unreal's code for making http requests. cesium-omniverse isn't doing any accessor-level caching, as far as I can tell. I've used CachingAssetAccessor in vsgCS with UrlAssetAccessor and it works, although I don't provide any user interface to cache size. I think the caching @r-veenstra is seeing when changing location is the result of the cesium-native tile selection algorithm, which doesn't unload tiles immediately that are out of view.

I think it's best to get #634 merged before adding the CachingAssetAccessor to cesium-omniverse.

timoore:
@r-veenstra What is the UI in cesium-unreal for the cache file and its maximum number of items?

Should these parameters be stored in Usd like the other tileset parameters? Probably, I would guess.

[...]

PR #666, opened 2024-02-05
Accessor cache using CeslumAsync::SqliteCache 

timoore:
Addresses #650.

This builds on #634, which needs to be merged first.


r-veenstra:
I did some benchmarking with this USDA
[load_test.zip](https://github.com/CesiumGS/cesium-omniverse/files/14178839/load_test.zip)

results:

First load, clear cache:  74 seconds, 116mb downloaded
Second load, cached: 44 seconds, 0mb downloaded
Restart app, load cached: 44 seconds, 0mb downloaded

Appears to be working as expected.

I do notice a rather large hang (~10-20 seconds) when I hit Reload all Tilesets in Cesium Debugging UI.  This isn't unique to this branch, and I suspect it has something to do with unloading all of the tileset resources before it begins reloaded.

Point is, if we ignore the hang, the actual loading of the cached data goes down to ~24 seconds.

r-veenstra:
Experimented here with the clear cache functionality, it appears to be working well!

See the results below - times with the red dash were after clearing cache, the rest are a reload from cached data.  Looking quite consistent.

image

r-veenstra:
Similar tests with Google Photorealistic 3D tiles

image

src/core/src/Context.cpp
```diff
@@ -145,6 +215,12 @@ void Context::reloadStage() {
     }
 }
 
+void Context::clearAccessorCache() {
+    if (_pCacheDatabase) {
+        _pCacheDatabase->clearAll();
```
lilleyse, line 220: Maybe I'm missing something, but I see this code is getting hit and no errors inside `clearAll`, but the `.sqlite` file remains the same size. Is that expected?

src/core/src/Context.cpp
```diff
@@ -31,21 +34,88 @@
...
+        if (maxCacheItems == 0) {
+            logger->oneTimeWarning("maxCacheItems set to 0, so disabling accessor cache");
+            return {};
+        } else if (auto dbName = getCacheDatabaseName(); !dbName.empty()) {
+            return std::make_shared<CesiumAsync::SqliteCache>(logger, getCacheDatabaseName(), maxCacheItems);
+        }
+        logger->oneTimeWarning("could not get name for cache database");
+        return {};
+    }();
+    return pCacheDatabase;
```
lilleyse, line 96: It seems like `maxCacheItems` will only get set on extension startup. Should it be possible for a user to set the value and have it update dynamically?

lilleyse (commented):
The caching functionality is working great, I just have some comments about cache location and settings.

src/core/src/Context.cpp
lilleyse, line 96: Is there an approach that avoids using static variables (i.e. global variables)? Ideally the cache database would be owned by `Context`.

src/public/CesiumOmniverse.cpp
```diff
@@ -185,6 +185,10 @@ class CesiumOmniversePlugin final : public ICesiumOmniverseInterface {
 #endif
     }
 
+    void clearAccessorCache() override {
+        _pContext->clearAccessorCache();
+    }
```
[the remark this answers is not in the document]
timoore: It wasn't clear to me that it it is `noexcept`; CesiumAsync::ICacheDatabase::clearAll() is not marked `noexcept`. I suppose that we could wrap a try / catch block inside.

src/core/src/Context.cpp
```diff
@@ -31,21 +34,88 @@
...
+        if (getResult == 0) {
+            homeDir = pwd.pw_dir;
+        }
+        delete[] buf;
+    }
+#elif defined(_WIN32)
+    if (char* cString = getenv("USERPROFILE")) {
+        homeDir = cString;
+    }
+#endif
```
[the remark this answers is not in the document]
timoore: I looked around a lot and found nothing. I meant to ask on the omniverse slack and think I did, but I will have to check. Other stuff like logs and config data are being written there.

src/core/src/Context.cpp
timoore: I did not expect that behavior, but I saw it too and checked the database with the sqlite command tool. All the rows are deleted by this call. I don't know enough about sqlite to say whether the file should be trimmed to or if that requires an additional operation.

src/core/src/Context.cpp
timoore: Probably. I was following cesium-unreal's implementation a bit to closely.

src/public/CesiumOmniverse.cpp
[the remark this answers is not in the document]
lilleyse: Hm, I was under the impression that everything in native was `noexcept` since it's expected to work in environments where extensions are disabled (e.g. Unreal). I wonder if that was an oversight or intentional.

In any case, the reason these functions are `noexcept` is because of [Carbonite plugin ABI rules](http://omniverse-docs.s3-website-us-east-1.amazonaws.com/carbonite/164.0-pre/docs/ABI.html#exceptions). CesiumOmniverse is meant to be a Carbonite plugin, though a few C++ runtime types have snuck into the interface, which should eventually be fixed (cc #3).

src/core/src/Context.cpp
timoore: That functionality -- updating the cache size after startup -- would have to be added to the cesium-native interface.

lilleyse: Ah ok. Maybe the UI just needs to say `(requires restart to take effect)` or something.

src/core/src/SettingsWrapper.cpp
```diff
@@ -12,6 +12,7 @@ const std::string_view SESSION_ION_SERVER_URL_BASE =
     "/persistent/exts/cesium.omniverse/sessions/session{}/ionServerUrl";
 const std::string_view SESSION_USER_ACCESS_TOKEN_BASE =
     "/persistent/exts/cesium.omniverse/sessions/session{}/userAccessToken";
+const char* MAX_CACHE_ITEMS_PATH = "/persistent/exts/cesium.omniverse/maxCacheItems";
```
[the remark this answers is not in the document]
timoore: It's passed directly to the carb::settings interface, which takes const char * values. The other settings go through fmt first. So const char * seems to me like the natural type for it; happy to change it if you feel otherwise.

lilleyse:
@timoore could you update `CHANGES.md`?

exts/cesium.powertools/cesium/powertools/utils/asset_cache_window.py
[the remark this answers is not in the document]
lilleyse: @r-veenstra bump

src/core/src/Context.cpp
lilleyse: @kring have you noticed this before in unreal/unity?

src/core/src/Context.cpp
lilleyse: @timoore bump on the UI change (any other suggestions welcome too)

lilleyse: @timoore bump - I think we should make this change.

src/core/include/cesium/omniverse/FilesystemUtil.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include <string>
+
+namespace cesium::omniverse::FilesystemUtil {
+std::string getCesiumCacheDirectory();
+std::string getUserHomeDirectory();
```
lilleyse, line 7: These should probably return `std::filesystem::path`

src/core/src/Context.cpp
```diff
@@ -33,19 +37,51 @@
 
 namespace cesium::omniverse {
 
+namespace {
+
+std::string getCacheDatabaseName() {
```
lilleyse, line 42: This could return `std::filesystem::path` as well

src/core/src/FilesystemUtil.cpp
```diff
@@ -0,0 +1,74 @@
...
+        }
+        delete[] buf;
+    }
+#elif defined(_WIN32)
+    if (char* cString = std::getenv("USERPROFILE")) {
+        homeDir = cString;
+    }
+#endif
+    return homeDir;
+}
```
lilleyse, line 73: It seems unlikely that `${omni_global_cache}` wouldn't exist. I would be ok with removing most of this backup code. It might also lead to some confusion since `${omni_global_cache}` is a different path than `$HOME/.nvidia-omniverse`.

src/core/src/Context.cpp
lilleyse: Sorry, I missed that this was addressed in https://github.com/CesiumGS/cesium-omniverse/commit/3308e8adddf9decabb858183ac60c389f2c1b791.

src/core/src/Context.cpp
kring: I haven't noticed it, but haven't looked either. It's not too surprising, I guess.

exts/cesium.powertools/cesium/powertools/utils/asset_cache_window.py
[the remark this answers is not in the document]
r-veenstra: @timoore @lilleyse I'm thinking a small settings button could live here:  
image

To get what i have in my image, replace line 191 in `main_window.py` with below

```
            with ui.HStack(height=24, spacing=3):
                self._profile_widget = CesiumOmniverseProfileWidget(self._cesium_omniverse_interface, height=20)
                self._add_button = ui.Button(
                    "",
                    image_url=f"{self._icon_path}/FontAwesome/plus-solid.png",
                    style=CesiumOmniverseUiStyles.setting_button_style,
                    enabled=True,
                    width=24
                )
```

And add this to `styles.py`

```

    setting_button_style = {
        "Button": {"padding": 0.0, "stack_direction": Direction.TOP_TO_BOTTOM, "margin": 0},
        "Button.Image": {
            "alignment": Alignment.CENTER,
        },
        "Button.Label": {"alignment": Alignment.CENTER_BOTTOM},
        "Button.Image:disabled": {"color": cl("#808080")},
        "Button.Label:disabled": {"color": cl("#808080")},
    }
```

The FA gear icon would be ok to use I think https://fontawesome.com/icons/gear?f=classic&s=solid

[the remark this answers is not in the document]
timoore: I have no problem with adding a settings button up there and having the cache parameters window evolve into a settings window. But are settings any less important than "Token", "Learn", and "Help"? What do you think about moving "Token" to the same settings window as the cache, and / or combining Learn, Help, and Settings into an  "other" menu?

lilleyse:
I added this to `CHANGES.md` in https://github.com/CesiumGS/cesium-omniverse/pull/696. So no need to update the change log in this PR.

lilleyse (approved):
Looks great, thanks @timoore

PR #668, opened 2024-02-06
Add a load timer powertool for load benchmarking

r-veenstra:
This PR adds a load timer powertool to assist with benchmarking load times for tilesets.

Rather than manually using a stopwatch or similar, this feature will automatically start a timer and trigger a reload of all tilesets in the stage.  The timer automatically completes once 30 successive frames of zero loading occurs (this assumes all loading has finished).

image

This could also be extended to load with and without cache once caching is added.

@timoore this should he helpful when testing load times in your libcurl/cache branch


corybarr (approved):
LGTM. I mentioned in an issue we might be able to in the future tap into an all-tiles-loaded event.

PR #670, opened 2024-02-07
add WmsRasterOverlay support and UI

### LINKED ISSUE #609: Support Web Map Service Raster Overlays
Cesium Native supports [Web Map Services (WMS)](https://github.com/CesiumGS/cesium-native/blob/8bfdcb39cf0738856cc9ad638d1eeef2c616f004/CesiumRasterOverlays/src/WebMapServiceRasterOverlay.cpp#L91) as raster overlays for tilesets.  With the work being done in https://github.com/CesiumGS/cesium-omniverse/pull/599, this implementation may now be easier.

image
_WMS Source containing Land Zoning information overlaid onto Cesium World Terrain._

The best current implementation of WMS is in Cesium for Unreal, look into the Cesium Web Map Service Raster Overlay component for more information.  

image

**Workflow**
- User has a tileset
- User has the ability to add a `CesiumWebMapServiceRasterOverlayPrim` as a child prim to the tileset (same concept as a regular ion imagery overlay)
- User enters the appropriate details for the WMS server in the UI (see User Interface below)
- WMS imagery loads in and is projected on top of the tileset.

**Other Requirements**
- Multiple stacked combinations of Cesium ion / WMS imagery should be supported (they should just stack in the same way as multiple Cesium ion layers stack now)
- WMS layers can have opacity, and should blend with layers below appropriately (see example)

image

**User Interface**
Use the Unreal Web Map Service Raster Overlay UI as a guide, it works well.

image

**Test data**

The following datasets can be used to test WMS overlays

NSW Land Zoning
Url: `https://mapprod3.environment.nsw.gov.au/arcgis/services/Planning/EPI_Primary_Planning_Layers/MapServer/WMSServer`
Layers: `4`
Georef: `-30.285129, 153.110419` (layers should be visible at this location)
image

Mercury
Url: `https://planetarymaps.usgs.gov/cgi-bin/mapserv?map=/maps/earth/moon_simp_cyl.map`
Layers: `LROC_WAC`
Georef: Anywhere

image

corybarr:
Adds Web Map Service (WMS) support:

image

Test file: [WmsTest.zip](https://github.com/CesiumGS/cesium-omniverse/files/14188362/WmsTest.zip)

Resolves #609


corybarr:
@r-veenstra I guess on good max ranges on the zoom level. Can you see if they're reasonable?

corybarr:
@r-veenstra I'm wondering if adding a "Refresh" button to the WMS raster overlay is the best way to handle users' property updates.

exts/cesium.usd.plugins/schemas/cesium_schemas.usda
```diff
@@ -512,6 +512,63 @@ class CesiumPolygonRasterOverlayPrim "CesiumPolygonRasterOverlayPrim" (
...
+
+    int cesium:maximumLevel = 14 (
+        customData = {
+            string apiName = "maximumLevel"
+        }
+        displayName = "Maximum Level"
+        doc = "Maximum zoom level."
+    )
+}
+
```
lilleyse, line 571: Just to double check, these are all standard / copied from unreal?

src/core/src/OmniWebMapServiceRasterOverlay.cpp
```diff
@@ -0,0 +1,102 @@
...
+    cesiumWebMapServiceRasterOverlay.GetLayersAttr().Get(&val);
+    return val;
+}
+
+void OmniWebMapServiceRasterOverlay::reload() {
+    const auto rasterOverlayName = UsdUtil::getName(_pContext->getUsdStage(), _path);
+
+    CesiumRasterOverlays::RasterOverlayOptions options;
+    options.showCreditsOnScreen = getShowCreditsOnScreen();
+    // options.ktx2TranscodeTargets = GltfUtil::getKtx2TranscodeTargets();
```
lilleyse, line 77: Any reason this is commented out?

src/core/src/OmniWebMapServiceRasterOverlay.cpp
```diff
@@ -0,0 +1,102 @@
...
+    options.loadErrorCallback = [this](const CesiumRasterOverlays::RasterOverlayLoadFailureDetails& error) {
+        _pContext->getLogger()->error(error.message);
+    };
+
+    CesiumRasterOverlays::WebMapServiceRasterOverlayOptions wmsOptions;
+    int minimumLevel = getMinimumLevel();
+    int maximumLevel = getMaximumLevel();
+    int tileWidth = getTileWidth();
+    int tileHeight = getTileHeight();
+    std::string layers = getLayers();
```
lilleyse, line 88: All of these could be `const auto`

src/core/src/OmniWebMapServiceRasterOverlay.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+    if (maximumLevel > minimumLevel) {
+        wmsOptions.minimumLevel = minimumLevel;
+        wmsOptions.maximumLevel = maximumLevel;
+    }
```
lilleyse, line 93: This doesn't look right...

src/core/src/UsdNotificationHandler.cpp
```diff
@@ -397,6 +398,39 @@ void processCesiumPolygonRasterOverlayChanged(
...
+    }
+
+    // Process base class first
+    processCesiumRasterOverlayChanged(context, webMapServiceRasterOverlayPath, properties);
+
+    auto reload = false;
+    auto updateBindings = false;
+
+    for (const auto& property : properties) {
+        if (property == pxr::CesiumTokens->cesiumBaseUrl || property == pxr::CesiumTokens->cesiumLayers) {
```
lilleyse, line 419: Should this also be checking `tileWidth`, `tileHeight`, `minimumLevel`, `maximumLevel`?

src/core/src/UsdNotificationHandler.cpp
```diff
@@ -551,6 +585,13 @@ void processCesiumPolygonRasterOverlayRemoved(Context& context, const pxr::SdfPa
     updateRasterOverlayBindings(context, polygonRasterOverlayPath);
 }
 
+void processCesiumWebMapServiceRasterOverlayRemoved(
+    Context& context,
+    const pxr::SdfPath& webMapServiceRasterOverlayPath) {
+    context.getAssetRegistry().removeWebMapServiceRasterOverlay(webMapServiceRasterOverlayPath);
+    updateRasterOverlayBindings(context, webMapServiceRasterOverlayPath);
+}
```
lilleyse, line 593: Make sure that when you merge in main that

```
    if (context.getAssetRegistry().getWebMapServiceRasterOverlay(webMapServiceRasterOverlayPath)) {
        return;
    }
```

is added

exts/cesium.usd.plugins/schemas/cesium_schemas.usda
```diff
@@ -455,7 +455,7 @@ class "CesiumRasterOverlayPrim" (
 }
 
 class CesiumIonRasterOverlayPrim "CesiumIonRasterOverlayPrim" (
-    doc = """Adds a prim for representing an Ion raster overlay. Should be a child of a tileset."""
+    doc = """Adds a prim for representing an Ion raster overlay."""
```
lilleyse, line 458: Thanks for fixing this

lilleyse (commented):
Looks great, just a couple small comments from me.

src/core/src/OmniWebMapServiceRasterOverlay.cpp
lilleyse: Actually it's fine. But I wonder if a warning should be logged if `maximumLevel < minimumLevel`

exts/cesium.usd.plugins/schemas/cesium_schemas.usda
corybarr: Correct. Everything down to the docs are a direct copy from Unreal.

src/core/src/OmniWebMapServiceRasterOverlay.cpp
corybarr: Shouldn't be there. I missed it looking at commits. WMS textures weren't loading so I did a quick check to see if it could possibly be transcoding.

src/core/src/OmniWebMapServiceRasterOverlay.cpp
corybarr: That's lifted from Unreal. I also thought about logging, then thought ideally it should be at the Python UI level. Users would have immediate and clear feedback. We should be able to add checking events into the DataModel when the values are changed. (The initial PR of tileset clipping did this.) I didn't want to crack that open in this PR, though. Now that I think about it, it shouldn't be hard. Let's try it at the Python level.

src/core/src/UsdNotificationHandler.cpp
corybarr: I'm hoping to get @r-veenstra's feedback on this one. With all the potential USD notifications from sliders, I'm concerned about spamming and too many reloadings. I'm wondering if the solution is to add a "Refresh" button or something to the WmsRasterOverlay.

exts/cesium.usd.plugins/schemas/cesium_schemas.usda
corybarr: "Ion" should be lowercase, right?

src/core/src/UsdNotificationHandler.cpp
corybarr: Added. I think it goes in `processCesiumWebMapServiceRasterOverlayAdded`

src/core/src/OmniWebMapServiceRasterOverlay.cpp
corybarr: This check was moved to the UI layer:
https://github.com/CesiumGS/cesium-omniverse/assets/1040194/3a80da3c-d165-4085-bea1-a33fe262601e

r-veenstra:
I've done some testing, seems to be working as expected.  Comments below.

- Clicking the Refresh Tiles button on a tileset or Reload All Tilesets in the Cesium Debug window doesn't result in the WMS overlay updating based on latest values - should it?

- We probably want to include these settings too:
image

src/core/src/UsdNotificationHandler.cpp
r-veenstra: A refresh button wouldn't be a bad idea, or ensuring Refresh Tileset button on the tileset also updates the overlays

r-veenstra:
> @r-veenstra I guess on good max ranges on the zoom level. Can you see if they're reasonable?

The defaults in there look okay and are consistent with Unreal too

src/core/src/UsdNotificationHandler.cpp
lilleyse: @corybarr I think we should still make this change in `UsdNotificationHandler` since the UI debounce issue is a broader issue.

corybarr:
> I've done some testing, seems to be working as expected. Comments below.
> 
> * Clicking the Refresh Tiles button on a tileset or Reload All Tilesets in the Cesium Debug window doesn't result in the WMS overlay updating based on latest values - should it?
> * We probably want to include these settings too:
>   image

@r-veenstra This is an issue with all raster overlays. I created #675 for this.

src/core/src/UsdNotificationHandler.cpp
corybarr: Sounds good. Let's go ahead and do that in this PR. I think there's some UX/UI we need to take a detailed look at before the next release. I created https://github.com/CesiumGS/cesium-omniverse/issues/676 to get it on the board.

corybarr: @lilleyse I added the check for these four properties.

Tangentially, the genericness of these tokens (`pxr::CesiumTokens->cesiumLayers`, `pxr::CesiumTokens->cesiumMinimumLevel`, etc) make me nervous. Would token-name collisions ever be an issue?

lilleyse: > Would token-name collisions ever be an issue?

I don't think so since it's context dependent. Maybe it would be a problem if a base class and sub class had the same token but I doubt the schema code generator would allow that.

corybarr:
> > Clicking the Refresh Tiles button on a tileset or Reload All Tilesets in the Cesium Debug window doesn't result in the WMS overlay updating based on latest values - should it?
> 
> Tilesets and raster overlays are now separate things in the C++ layer so `OmniTileset` would need to explicitly reload the `OmniRasterOverlays` that it references. @corybarr could you make that fix?

@lilleyse Added.

lilleyse (approved):

PR #673, opened 2024-02-09
Fix crash on extension reload

lilleyse:
Creating a Fabric prim with the same path as a previously destroyed prim causes a crash. Previously this was avoided by included the object id and pool id in the name, but on extension reload the pool id is initialized to 0 and new paths may conflict with old paths.

Now a contextId is randomly generated and ensures that Fabric prim paths are unique even across extension reloads.

There are probably more robust and safer ways to do this than `glm::linearRand<int64_t>(0, 100000)`, like using a timestamp, but since extension reloads are uncommon and I didn't want to overcomplicate this.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 * Fixed zooming to tileset extents when tileset prims have non identity transformation.
 * Fixed crash when updating tilesets shader inputs.
 * Fixed crash when setting certain `/Cesium` debug options at runtime.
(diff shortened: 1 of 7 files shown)


corybarr (approved):
Simple and clean.

My goto-for these issues has been a timestamp like you mention in a one-liner. I don't think it would add any real amount of complexity if you were to drop it in. But the odds of a name-collision crash are also insignificant.

lilleyse:
I switched to a timestamp; it's a better solution.

Issue #198, opened 2023-03-02
Pause image / video capture until stage has finished loading


r-veenstra:
When capturing video using Omniverse Create's Movie Capture tool, it does not wait for our tilesets to complete loading before it captures a frame.  The result is flickering data, or completely unloaded sections of data in videos.

It would be great if we could somehow notify Omniverse to wait until our data has loaded, before proceeding to capture a frame.  No idea if this is actually feasible, but it is desireable.

Currently the workaround is to nominate an arbitary "Settle" time, where Omniverse renders XX frames before doing another capture.  This has the effect of delaying frame capture, allowing time for loads to complete, however the amount of time nominated is a complete guess and there is no guarantee it would be sufficient for the length of a capture

PR #674, opened 2024-02-12
Wait for loading tiles when using the movie capture tool

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/198.

I confirmed that `updateViewOffline` is called when the movie capture tool is active, and `updateView` is called otherwise, but I haven't been able to verify the results because I'm having trouble with the [Capturing Images and Video](https://cesium.com/learn/omniverse/omniverse-capturing-images/) tutorial. Marking as draft until that's resolved.


r-veenstra:
Looks good.  Both these are with settle latency at 0


**0.17.0:**

https://github.com/CesiumGS/cesium-omniverse/assets/123468416/01fccd05-b5c6-43ec-a304-4e8db3ef3dd7

**This branch:**

https://github.com/CesiumGS/cesium-omniverse/assets/123468416/7e8997e9-f17d-4c79-9b0d-ebe90996b0e8

Captures are much faster than when using an arbitary value for settle latency too!

lilleyse:
Great, I'll mark this as ready for review then

corybarr:
Is the changing geometry in the render from this branch due to geometric error?

corybarr (approved):

lilleyse:
> Is the changing geometry in the render from this branch due to geometric error?

I believe so

PR #680, opened 2024-02-12
Fix crash when dependent prims are removed in specific order

lilleyse:
The following scenario would cause a crash:

* Receive notification that CesiumIonServer was removed
* Receive notification that CesiumIonRasterOverlay was removed

When prims are removed they update any assets that have a relationship to them. If the other assets had also been removed from the stage, during their update they may try to access properties on a USD prim that no longer exists.

I ended up making the fix in UsdNotificationHandler so that it checks whether prims exist before updating them.

I was able to trigger this by loading [imagery-single-2.zip](https://github.com/CesiumGS/cesium-omniverse/files/14251881/imagery-single-2.zip), hand deleting the raster overlay and rel from the .usda, and clicking the fetch changes button. This USD is a bit older and doesn't have an ion server in it, so the default ion server gets removed and the raster overlay gets removed.

### DIFF
--- a/src/core/src/UsdUtil.cpp
+++ b/src/core/src/UsdUtil.cpp
@@ -152,12 +152,12 @@ pxr::GfMatrix4d glmToUsdMatrix(const glm::dmat4& matrix) {
 
 glm::dmat4 computePrimLocalToWorldTransform(const pxr::UsdStageWeakPtr& pStage, const pxr::SdfPath& path) {
     const auto prim = pStage->GetPrimAtPath(path);
+    const auto xform = pxr::UsdGeomXformable(prim);
 
-    if (!prim.IsA<pxr::UsdGeomXformable>()) {
+    if (!isSchemaValid(xform)) {
         return glm::dmat4(1.0);
     }
 
-    const auto xform = pxr::UsdGeomXformable(prim);
     const auto time = pxr::UsdTimeCode::Default();
     const auto transform = xform.ComputeLocalToWorldTransform(time);
     return usdToGlmMatrix(transform);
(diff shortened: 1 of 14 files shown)


corybarr (approved):

src/core/src/UsdUtil.cpp
```diff
@@ -246,7 +246,7 @@ bool primExists(const pxr::UsdStageWeakPtr& pStage, const pxr::SdfPath& path) {
 }
 
 bool isSchemaValid(const pxr::UsdSchemaBase& schema) {
-    return schema.GetSchemaKind() != pxr::UsdSchemaKind::Invalid;
+    return schema.GetPrim().IsValid() && schema.GetSchemaKind() != pxr::UsdSchemaKind::Invalid;
```
corybarr, line 249: I can see this working, but I'm wondering if it changes the intent of the method. Most of the calls are `!UsdUtil::isSchemaValid(cesiumTileset)`, where we seem to be checking for prim existence. Why are we checking for schema validity instead of prim existence?

lilleyse: I think it's worth checking both in case the schema is created from a prim with the wrong type. I'm not sure when that would actually happen, but the extra safeguard doesn't hurt.

corybarr: True. I think a better way to ask is "What do we get from checking schema validity vs checking for prim existence across the code base?" Could it all just be a call to [Usd.Object.IsValid()](https://docs.omniverse.nvidia.com/dev-guide/latest/programmer_ref/usd/prims/check-prim-exists.html)? `!UsdUtil::isSchemaValid(cesiumTileset)` makes me wonder if there's something particularly special about the schema that needs to be checked. Or is `isSchemaValid` a more robust version of `primExists`?

lilleyse: > Or is isSchemaValid a more robust version of primExists?

That's how I see it, since a schema wraps a prim.

Issue #676, opened 2024-02-12
Spec and implement improved UI/UX for refreshing and reloading tilesets and raster overlays


corybarr:
Currently, some edits to raster overlays, cartographic polygons, etc, can result in a cascade of overloads. Starting from a product level, we should revisit the UX/UI with an eye out for the following:

- Avoiding tileset reloads
- Avoiding unused data request to external data sources
- If and when a user should request tileset and raster-overlay refreshes
- Potential debouncing techniques

Cesium for Unreal currently has some different approaches than Cesium for Omniverse. Listing these differences and deciding if they should be adopted provides a starting point.

lilleyse:
One idea for cutting down on tileset reloads is to hardcode a maximum number of raster overlays per tile so that the material doesn't need to be recreated each time a new raster overlay is added or removed.

corybarr:
I also wonder how feasible it is to add the visibility-toggle eye icon that USD prims have for each raster overlay. Changes to non-visible raster overlays could be batched until visible. Aside from being a way to disable reload, it also might make data exploration like comparison between raster-overlay datasets easier.

Issue #368, opened 2023-06-26
Guide for working with Cesium USD schemas


lilleyse:
There should be a guide for working with Cesium USD schemas. A few things that could be covered:

**General**

* The cesium schemas are located in `cesium_schemas.usda`. In general this is the only file you need to edit.
* Files in `src/plugins/CesiumUsdSchemas/src/CesiumUsdSchemas` are autogenerated except `module.cpp` and `moduleDeps.cpp`

**Modifying Schemas**

* Any time you add, remove, or modify an attribute in  `cesium_schemas.usda` make sure to:
  * Run `regenerate_schema.sh` (or `.bat` in windows). This will update the autogenerate files.
  * Run `genStubs.sh` (or `.bat` in windows). This will regenerate Python stubs. Note that `genStubs` only works in Kit 105.
  * Update Omniverse UI. E.g. see `tileset_attributes_widget.py`
  * Update `Context.cpp` for `TfNotice` handling. E.g. see `processCesiumTilesetChanged`
* Any time you add a class the process is slightly more involved:
  * Run `regenerate_schema.sh`
  * Add the new `[class].cpp` to `SOURCES` in `CesiumUsdSchemas/CMakeLists.txt`
  * Add the new `wrap[Class].cpp` to `PY_BINDING_SOURCES` in `CesiumUsdSchemas/CMakeLists.txt`
  * Update various parts of the C++ and Python code (see what other classes are doing)
  * Run `genStubs.sh`
* If the class is inheriting from a pxr class other than `Typed` or `APISchemaBase`:
  * Add the library to `subLayers` in `cesium_schemas.usda`
  * Add the library to `CesiumUsdSchemas/CMakeLists.txt` in the `setup_lib` call
  * Add the library to `moduleDeps.cpp`
  * See `usdGeom` for reference

PR #681, opened 2024-02-13
Add additional raster overlay prim properties and UI

### LINKED ISSUE #675: Add settings for screen-space error, max texture size, etc to raster overlays
Unreal currently exposes additional options for controlling raster overlays that aren't exposed in Omniverse:

image

Some of these settings (including max SSE, max tile downloads) are on the tile-set level in Omniverse. Proposal: match Unreal's approach by moving these settings from the tileset and adding the additional ones.

corybarr:
Adds the following to all raster overlay prims:

image

Resolves #675


exts/cesium.omniverse/cesium/omniverse/ui/attributes/ion_raster_overlay_attributes_widget.py
```diff
@@ -25,11 +25,8 @@ def _customize_props_layout(self, props):
                 CustomLayoutProperty("cesium:ionAssetId")
                 CustomLayoutProperty("cesium:ionAccessToken")
                 CustomLayoutProperty("cesium:ionServerBinding")
-            with CustomLayoutGroup("Rendering"):
-                CustomLayoutProperty("cesium:alpha", build_fn=build_slider(0, 1))
                 CustomLayoutProperty("cesium:overlayRenderMethod")
```
lilleyse, line 28: Should `cesium:overlayRenderMethod` be added to `build_common_raster_overlay_properties` and removed from here?

exts/cesium.omniverse/cesium/omniverse/ui/attributes/custom_attribute_widgets.py
```diff
@@ -105,3 +106,18 @@ def custom_slider(stage, attr_name, metadata, property_type, prim_paths, *args,
...
+    with CustomLayoutGroup("Maximum Screen Space Error"):
+        CustomLayoutProperty("cesium:maximumScreenSpaceError")
+    with CustomLayoutGroup("Maximum Texture Size"):
+        CustomLayoutProperty("cesium:maximumTextureSize")
+    with CustomLayoutGroup("Maximum Simultaneous Tile Loads"):
+        CustomLayoutProperty("cesium:maximumSimultaneousTileLoads")
+    with CustomLayoutGroup("Sub Tile Cache Bytes"):
+        CustomLayoutProperty("cesium:subTileCacheBytes")
+    with CustomLayoutGroup("Credit Display"):
+        CustomLayoutProperty("cesium:showCreditsOnScreen")
```
lilleyse, line 123: Maybe this file should be renamed to `attribute_widget_util.py`?

src/core/src/UsdNotificationHandler.cpp
```diff
@@ -332,6 +339,11 @@ void processCesiumRasterOverlayChanged(
     if (updateRasterOverlayAlpha) {
         updateRasterOverlayBindingsAlpha(context, rasterOverlayPath);
     }
+
+    if (updateRasterOverlayOptions) {
+        auto pNativeOverlay = pRasterOverlay->getRasterOverlay();
+        pRasterOverlay->updateRasterOverlayOptions(pNativeOverlay->getOptions());
```
lilleyse, line 345: Even though `getRasterOverlay` is a public function it doesn't feel right that `UsdNotificationHandler` is accessing the native object. What about restructuring `OmniRasterOverlay` so the functions look like this:

```cpp
CesiumRasterOverlays::RasterOverlayOptions OmniRasterOverlay::createRasterOverlayOptions() const {
    CesiumRasterOverlays::RasterOverlayOptions options;
    options.ktx2TranscodeTargets = GltfUtil::getKtx2TranscodeTargets();
    setRasterOverlayOptionsFromUsd(options);
    return options;
}

void OmniRasterOverlay::updateRasterOverlayOptions() {
    const auto pRasterOverlay = getRasterOverlay();
    if (pRasterOverlay) {
        setRasterOverlayOptionsFromUsd(pRasterOverlay->getOptions());
    }
}

void OmniRasterOverlay::setRasterOverlayOptionsFromUsd(CesiumRasterOverlays::RasterOverlayOptions& options) const {
    options.showCreditsOnScreen = getShowCreditsOnScreen();
    options.maximumScreenSpaceError = getMaximumScreenSpaceError();
    options.maximumTextureSize = getMaximumTextureSize();
    options.maximumSimultaneousTileLoads = getMaximumSimultaneousTileLoads();
    options.subTileCacheBytes = getSubTileCacheBytes();
}
```

Previously:

```cpp
CesiumRasterOverlays::RasterOverlayOptions OmniRasterOverlay::getUsdOptions() {
    CesiumRasterOverlays::RasterOverlayOptions options;
    setOptionsFromUsd(options);
    return options;
}

void OmniRasterOverlay::updateRasterOverlayOptions(CesiumRasterOverlays::RasterOverlayOptions& options) {
    setOptionsFromUsd(options);
}

void OmniRasterOverlay::setOptionsFromUsd(CesiumRasterOverlays::RasterOverlayOptions& options) {
    options.showCreditsOnScreen = getShowCreditsOnScreen();
    options.ktx2TranscodeTargets = GltfUtil::getKtx2TranscodeTargets();
    options.maximumScreenSpaceError = getMaximumScreenSpaceError();
    options.maximumTextureSize = getMaximumTextureSize();
    options.maximumSimultaneousTileLoads = getMaximumSimultaneousTileLoads();
    options.subTileCacheBytes = getSubTileCacheBytes();
}
```

src/core/src/OmniRasterOverlay.cpp
```diff
@@ -38,6 +39,32 @@ double OmniRasterOverlay::getAlpha() const {
...
+    int value;
+    cesiumRasterOverlay.GetMaximumSimultaneousTileLoadsAttr().Get(&value);
+    return static_cast<int>(value);
+}
+int OmniRasterOverlay::getSubTileCacheBytes() const {
+    const auto cesiumRasterOverlay = UsdUtil::getCesiumRasterOverlay(_pContext->getUsdStage(), _path);
+    int value;
+    cesiumRasterOverlay.GetSubTileCacheBytesAttr().Get(&value);
+    return static_cast<int>(value);
+}
```
lilleyse, line 66: These will need to be updated to check `CesiumUtil::isSchemaValid(cesiumRasterOverlay)` once https://github.com/CesiumGS/cesium-omniverse/pull/680 is merged.

exts/cesium.omniverse/cesium/omniverse/ui/attributes/custom_attribute_widgets.py
```diff
@@ -105,3 +106,18 @@ def custom_slider(stage, attr_name, metadata, property_type, prim_paths, *args,
         )
 
     return custom_slider
+
+
+def build_common_raster_overlay_properties():
```
lilleyse, line 111: It's starting to feel like there are too many groups with just a single option. Can they be consolidated? I guess copy whatever Unreal is doing.

image

lilleyse (commented):
Looks good, I left some comments. I'll test it out after the next round of changes is in.

exts/cesium.usd.plugins/schemas/cesium_schemas.usda
```diff
@@ -452,6 +452,38 @@ class "CesiumRasterOverlayPrim" (
         displayName = "Overlay Render Method"
         doc = "The Cesium default material will give the raster overlay a different rendering treatment based on this selection."
     )
+
+    uniform float cesium:maximumScreenSpaceError = 2.0 (
```
lilleyse, line 456: Could you update `CHANGES.md` for this PR and https://github.com/CesiumGS/cesium-omniverse/pull/670?

* Added support for WMS raster overlays
* Added raster overlay options: `maximumScreenSpaceError `, `maximumTextureSize`, `maximumSimultaneousTileLoads`, `subTileCacheBytes`

exts/cesium.omniverse/cesium/omniverse/ui/attributes/ion_raster_overlay_attributes_widget.py
corybarr: Thinking about it...yes. I moved it there. It's only in `ion_raster_overlay_attributes_widget` because it's currently less than useful for polygon raster overlays and probably will never make sense for WMS and related raster overlays.

So I added a function param where it's suppressed by default. To get it in the widget, you need to do this:

```
        with frame:
            ...
            build_common_raster_overlay_properties(add_overlay_render_method=True)
```
It's still of course under "Raw USD Properties." I'm not sure if that can be suppressed. But it being only there hopefully communicates that it's not a property we're currently handling.

exts/cesium.omniverse/cesium/omniverse/ui/attributes/custom_attribute_widgets.py
corybarr: Sure, we can do that. 

I left the old name because "util" essentially means "stuff." I try to avoid it, even though it's often unavoidable. NVIDIA has UsdPropertiesWidgetBuilder to help get away from the default, so maybe `cesium_properties_widget_builder.py` ultimately make sense--but it's not a class. "util" is fine for now. Maybe once the UI grows, the commonality of what's there will inform a more precise name.

exts/cesium.omniverse/cesium/omniverse/ui/attributes/custom_attribute_widgets.py
corybarr: I agree. I took a stab at this maybe two months ago, but didn't find a huge improvement. Here's what I found. Looking again, I think we can improve the UI but can't match Unreal. 

Here's a comparison of the ion raster overlay UIs across frameworks. The OV interface has been tweaked to what I think is best:

image

The core limitation is that in Omniverse, everything has to be in a collapsible group. (At least I haven't found a way around this.) If you don't put a `LayoutProperty` in a `LayoutGroup`, OV will do it for you. In the image above, OV is providing "Extra Properties." We can replace that default string, but can't de-group it like in Unreal.

Looking back, I think "Extra Properties" is better than each isolated `LayoutProperty` having it's own redundant `LayoutGroup`. It's possible there's a way to customize `CustomLayoutFrame`, but I'm not seeing any easy solution.

src/core/src/UsdNotificationHandler.cpp
corybarr: Yes, I like that better. 

I think `updateRasterOverlayOptions` should also be `const`.

I wonder if it makes sense to move `OmniRasterOverlay::getRasterOverlay()` from `public` to `protected`? The only line that calls it is `OmniTileset` in the code below. We could `friend` the `add`. It seems nice to enforce Omni* classes only talking to other Omni* classes. Then the only way to get to a Native object is via its Omni class.

```
    for (const auto& boundRasterOverlayPath : boundRasterOverlayPaths) {
        const auto pOmniRasterOverlay = _pContext->getAssetRegistry().getRasterOverlay(boundRasterOverlayPath);
        if (pOmniRasterOverlay) {
            pOmniRasterOverlay->reload();
            const auto pNativeRasterOverlay = pOmniRasterOverlay->getRasterOverlay();
            if (pNativeRasterOverlay) {
                _pTileset->getOverlays().add(pNativeRasterOverlay);
            }
        }
    }
```

exts/cesium.omniverse/cesium/omniverse/ui/attributes/custom_attribute_widgets.py
lilleyse: "builder" sounds good

src/core/src/UsdNotificationHandler.cpp
lilleyse: Sure, a friend class works. I don't tend to use them but I don't have an issue with it.

src/core/src/OmniRasterOverlay.cpp
lilleyse: Alright, https://github.com/CesiumGS/cesium-omniverse/pull/680 was merged

corybarr: Updated.

I'm not crazy about the required sync of default values between these accessors and the schema file:

```
int OmniRasterOverlay::getSubTileCacheBytes() const {
    const auto cesiumRasterOverlay = UsdUtil::getCesiumRasterOverlay(_pContext->getUsdStage(), _path);
    if (!UsdUtil::isSchemaValid(cesiumRasterOverlay)) {
        return 16777216;
    }
```

I don't necessarily have a better solution. I pushed the sync responsibility to a bullet in #368. It might be possible to not rely on documentation if we create a `GetDefaultAttr(prim, attrName)` function. This would look nice on the surface, but I can't think of better way than to define messy scratch prims to query. And I'm not even sure if they're immediately queryable.

src/core/src/UsdNotificationHandler.cpp
corybarr: I added two friend methods instead of a class so `OmniTileset` wouldn't have access to all the `OmniRasterOverlay` internals. The friend class would be a clean one-liner but less locked down.

src/core/src/OmniRasterOverlay.cpp
lilleyse: I don't like the default values either...

Maybe the default values should be grabbed from a scratch `RasterOverlayOptions` / `TilesetOptions`? This doesn't work for all files though.

lilleyse (approved):

src/core/src/UsdNotificationHandler.cpp
```diff
@@ -332,6 +339,10 @@ void processCesiumRasterOverlayChanged(
     if (updateRasterOverlayAlpha) {
         updateRasterOverlayBindingsAlpha(context, rasterOverlayPath);
     }
+
+    if (updateRasterOverlayOptions) {
+        pRasterOverlay->updateRasterOverlayOptions();
+    }
```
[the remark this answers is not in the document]
corybarr: I'm not positive. I created #676 largely in response to how reloads for WebMapServiceRasterOverlayOptions should happen. 

I believe Unreal expects users to manually refresh to get updated data for WMS but also other raster overlays. I very much don't like how WebMapServiceRasterOverlayOptions in OV triggers a reload and potential server spam. I think we should look closely at this UX before the March release.

Issue #638, opened 2024-01-26
Cesium Window shows "Connect to Cesium ion" after reloading a stage


corybarr:
To reproduce:

1. Open a stage
2. Reload the stage
3. Check the Cesium window

It seems like this is a UI issue, not an actual authentication issue. Assets still load:

![image](https://github.com/CesiumGS/cesium-omniverse/assets/1040194/5de19ac9-615f-4f95-b624-23631dcbd8ac)

PR #683, opened 2024-02-13
Resume ion session on reload

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/638


corybarr (approved):
Most impactful one-line PR in a long time.

PR #684, opened 2024-02-13
Remove deprecated properties from CesiumDataPrim

### LINKED ISSUE #598: Remove deprecated USD properties and single ion server backwards compatibility
Previously, tilesets and imagery used the default ion server and default access token in `CesiumDataPrim`. With support added for multiple ion servers, tilesets and imagery now reference a `CesiumIonServerPrim`. We have backwards compatibility for older USD files but eventually that should be removed.

That includes:

* Removing `cesium:projectDefaultIonAccessToken` and `cesium:projectDefaultIonAccessTokenId` from `CesiumDataPrim`
* Removing backwards compatibility code in `extension.py` [here](https://github.com/CesiumGS/cesium-omniverse/blob/ccccd26a17f28c82fa18fa17dcb6415d5a4b6381/exts/cesium.omniverse/cesium/omniverse/extension.py#L471-L476)
* Remove backwards compatibility code in `Context.cpp` [here](https://github.com/CesiumGS/cesium-omniverse/blob/fb181ccbab57d371da9a5a9cc9e6b10739ada315/src/core/src/Context.cpp#L179-L199)

lilleyse:
Fixes #598


exts/cesium.usd.plugins/cesium/usd/plugins/CesiumUsdSchemas/__init__.pyi
```diff
@@ -570,7 +559,9 @@ class WebMapServiceRasterOverlay(RasterOverlay):
     @classmethod
     def GetSchemaAttributeNames(cls, *args, **kwargs) -> Any: ...
     @classmethod
-    def GetWidthAttr(cls, *args, **kwargs) -> Any: ...
+    def GetTileHeightAttr(cls, *args, **kwargs) -> Any: ...
+    @classmethod
+    def GetTileWidthAttr(cls, *args, **kwargs) -> Any: ...
```
corybarr, line 564: Apologies. Looks like I didn't commit some updated bindings.

exts/cesium.omniverse/cesium/omniverse/bindings/CesiumOmniversePythonBindings.pyi
```diff
@@ -71,7 +71,7 @@ class ICesiumOmniverseInterface:
     def credits_start_next_frame(self) -> None: ...
     def get_asset_token_troubleshooting_details(self, *args, **kwargs) -> Any: ...
     def get_asset_troubleshooting_details(self, *args, **kwargs) -> Any: ...
-    def get_credits(self) -> List[Tuple[str, bool]]: ...
+    def get_credits(self) -> List[Tuple[str,bool]]: ...
```
corybarr, line 74: This is exactly what's failing in CI. There seems to be discrepancies with black in CI and in dev environments.

```
-    def get_credits(self) -> List[Tuple[str,bool]]: ...
+    def get_credits(self) -> List[Tuple[str, bool]]: ...
```

corybarr (approved):

exts/cesium.omniverse/cesium/omniverse/bindings/CesiumOmniversePythonBindings.pyi
lilleyse: Should be fixed in https://github.com/CesiumGS/cesium-omniverse/pull/684/commits/9cd8a8305be4f9b068b53913d247e6f481bd94eb. I had to mess with the `genStubs` script a bit to run `black` properly.

PR #691, opened 2024-02-16
Add Tile Map Service (TMS) and Web Map Tile Service (WMTS) raster overlay support

corybarr:
**(Draft until UI polishing is complete)**

*********************************************

Adds TMS support:

image

Adds WMTS support:

image

- WMTS Test USDA: [WmtsTest.zip](https://github.com/CesiumGS/cesium-omniverse/files/14315931/WmtsTest.zip)
- TMS Test USDA:  [TmsTest.zip](https://github.com/CesiumGS/cesium-omniverse/files/14315946/TmsTest.zip)

*************************************************

**UI/UX Notes:**

* If a parameter is not relevant based on another set value, Unreal will gray those options out:

image

This PR hides them instead. I didn't find an obvious way to gray them out.

* In Unreal, "Tile Matrix Set Labels" is an array of strings:
image

While Omniverse supports string[] parameters, they aren't by default editable in the UI: 
image

In this PR, users supply the strings comma-separated.

* I don't have an example to test all settings. "Tiling Scheme Settings" is currently untested: 
image

* I hit some limitations trying to visually group parameters in the custom attribute widgets. In particular omni.ui.Line didn't appear. The UI is functional, but the UX and in particular visual groupings could use improvement.

* reloading the tileset every time a property is changed is probably not ideal. See #676


corybarr:
> This appears to be functioning as expected.
> 
> My only question is should there be some feedback in the console when WMTS is configured incorrectly?
> 
> For example, with WMS and TMS if I put in a wrong URL, I'll usually get some form of error.
> 
> However, with WMTS nothing displays in the log at all.

I think it would be better. I'll look into it. . 

For ref, it seems that a valid, although empty, result is returned when WMTS Urls are bad:
image

A bad TMS Url will fail the `if (result)` test and trigger the error logger.

PR #693, opened 2024-02-22
Fix token window

lilleyse:
Fixes a bug in https://github.com/CesiumGS/cesium-omniverse/pull/684 where the token window was trying to access the project default access token from the `/Cesium` prim instead of the active server.

image


corybarr (approved):
LGTM

Issue #695, opened 2024-02-23
ion authentication fails for packaged build


lilleyse:
ion authentication is failing for packaged builds on Linux after https://github.com/CesiumGS/cesium-omniverse/pull/634 was merged.

![image](https://github.com/CesiumGS/cesium-omniverse/assets/915398/46e1ddda-70a5-4f57-9d62-2a09a4dc1b52)

Notes so far:

* Only affects packaged builds i.e. those built from the AlmaLinux8 docker container
* Doesn't affect local builds
* Updating curl, OpenSSL, etc to the newest versions doesn't fix it
* A workaround would be to ship our own cert again

Testing steps:

* Switch to [update-libraries](https://github.com/CesiumGS/cesium-omniverse/tree/update-libraries) which has some Docker file fixes
* Run `scripts/build_package_almalinux8.sh`. This produces a zip folder in the build folder.
* Extract the zip file and copy the extensions to `/home/slilley/Documents/Kit/shared/exts/`
* Run USD Composer and sign in to Cesium ion

lilleyse:
It turns out that the OpenSSL search dir is hardcoded into the binary. In CentOS/RHEL/Alma that path is `/etc/pki/tls` whereas on Ubuntu it's `/etc/ssl/certs/`. So an Alma build of OpenSSL won't be able to locate certs in Ubuntu.

There are a few options

* Override OPENSSLDIR when compiling OpenSSL (but now that breaks Omniverse users on CentOS/RHEL)
* Find some portable way to locate OS certificate dir at compile time or runtime
* Ship our own cert again

@weegeekps any thoughts? You've dealt with these sorts of issues before...

weegeekps:
> You've dealt with these sorts of issues before...

Yup, and these are the exact issues as to why the Curl devs recommend you just bundle your own certs, including [providing an mirror for the Mozilla ca-cert bundle](https://curl.se/docs/caextract.html).

The ultimate problem here is that each OS and distribution uses their own paths, and most also occasionally change the paths between major versions and even OpenSSL versions. Debian/Ubuntu is particularly bad about this. You can try to do some heuristics to check the various known paths for the major OS's and their variants, but ultimately you will inevitably run into problems.

I can't find exactly where it's kept in Omniverse but I believe Nvidia also ships the Mozilla ca-cert bundles as well.

[...]

PR #697, opened 2024-02-26
Revert to using our own SSL certificate

timoore:
The certificates do not live in the same directory in different Linux distributions, one of the many reasons why it's not portable to rely on the system's installed certificates. Addresses #695 .


lilleyse (approved):
I don't have my usual dev machine to test with, but this looks good.

r-veenstra:
I've tested with several cesium-omniverse-samples and it all appears to be working well for me

PR #699, opened 2024-02-29
Remove unused packages in CI to avoid low disk space errors

r-veenstra:
Linux runners in CI were encountering low disk space errors, causing CI to fail.

This PR cleans up unused packages already installed on the linux images, clearing approx 4GB of space.  

CI builds are now successful.

This is likely a bandaid and we may encounter it again in the future, however it is necessary in order to get the next release out.


timoore (approved):
Looks good to me. Mercurial?!

PR #706, opened 2024-03-27
Add scrollbar to Main UI and Powertools UI

### LINKED ISSUE #705: Main Cesium UI is missing scrollbar
When the Cesium for Omniverse extension window is visible in a low resolution environment, or just with a small sidebar, items such as the sign in button or the quick add options can't be accessed without making the window larger.  Adding a simple vertical scrollbar will solve this.

r-veenstra:
This resolves #705 by adding a vertical scrollbar to the main window and also powertools window.

Similar to Cesium for Unreal, I've made only the UI below the main button scrollable

image

Due to the way `ui.ScrollingFrame()` works by default, there will be a small gap present when the scrollbar isn't visible, as opposed to the contents responsively resizing to fill the whole window.  

image

This appears to be common behaviour across Omniverse UI, so I'm not too concerned about this.

image


lilleyse (commented):
Awesome, thanks @r-veenstra

lilleyse (approved):

PR #709, opened 2024-04-12
Update conan version in developer-setup README

jshrake-nvidia:
Update conan to 1.64.0 because 1.60.0 does not work correctly with Python 3.12 (see https://github.com/conan-io/conan/issues/14801).


lilleyse:
Thanks @jshrake-nvidia, I'll follow up with another PR that updates the docker image.

PR #710, opened 2024-04-12
Use Conan 1.64.0 in CI and docker

lilleyse:
Follow up to https://github.com/CesiumGS/cesium-omniverse/pull/709


lilleyse:
I switched to Conan 1.63.0 because 1.64.0 isn't available yet in choco

PR #711, opened 2024-04-29
Fix missing button for adding imagery from Asset UI

### LINKED ISSUE #701: Use as Terrain Tileset Base Layer button missing
The Use as Terrain Tileset Base Layer button in the Assets UI appears to be missing for Imagery assets.

https://community.cesium.com/t/geotiff-data-can-no-longer-be-linked-to-omniverse/30363

An initial test reproduction:

- Add Cesium World Terrain to stage using the Add to Stage button in the assets UI
- Select an Imagery asset in the assets UI - No button appears for adding the layer

r-veenstra:
Closes #701

The Use as Terrain Tileset Base Layer button was missing from the assets UI when selecting an imagery asset to add

Two issues were present

Firstly, the `self._asset_type` needed to be checking for `IMAGERY` when checking if the button should be created in `asset_details_widget.py`

Secondly, `is_tileset `was being passed a prim instead of a prim path, which caused the function to fail

The button is now visible, and successfully adds to either the first tileset found in the selection, the first tileset in the stage if no selection is present, or creates a new Cesium World Terrain tileset if no tilesets are found.

image

### DIFF
--- a/exts/cesium.omniverse/cesium/omniverse/ui/asset_details_widget.py
+++ b/exts/cesium.omniverse/cesium/omniverse/ui/asset_details_widget.py
@@ -76,7 +76,7 @@ class CesiumAssetDetailsWidget(ui.ScrollingFrame):
         selection = context.get_selection().get_selected_prim_paths()
         tileset_path: Optional[str] = None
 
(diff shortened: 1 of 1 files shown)


lilleyse:
Thanks @r-veenstra

PR #712, opened 2024-04-29
Update cesium-native

lilleyse:
Updating cesium-native to bring in this fix: https://github.com/CesiumGS/cesium-native/pull/868


lilleyse:
This commit is slightly behind head - we need more substantial changes before we can incorporate https://github.com/CesiumGS/cesium-native/pull/841

PR #714, opened 2024-05-07
Added `ecefToUsdTransform` to `CesiumGeoreferencePrim`

lilleyse:
* Added read-only attribute `ecefToUsdTransform` to `CesiumGeoreferencePrim`. Previously this was stored in `/CesiumSession` which has since been removed (and this PR scrubs it out of the code base)
* Fixed crash when updating globe anchor when georeferencing is disabled.


lilleyse:
@r-veenstra this PR includes USD schema changes, so if you're doing the June release make sure to update the `cesium.usd.plugins` version number

Issue #432, opened 2023-08-16
Add initial point-cloud UI with controls for base size, max attenuation, etc.


corybarr:
Presently there are no user-facing controls for extension users to edit the size and other parameters of point-cloud data to fit their needs. We need to add a UI, similar to Unity. The initial controls should allow control of the base point size at least.

corybarr:
For reference, these are the controls for Cesium for Unity:

![image](https://github.com/CesiumGS/cesium-omniverse/assets/1040194/46f34c30-9341-49ea-ab2e-f39d6f45771a)

PR #718, opened 2024-05-13
Added pointSize attribute

lilleyse:
Added `pointSize` attribute to `CesiumTilesetPrim` for controlling the size of points in meters. This is just a small step towards addressing https://github.com/CesiumGS/cesium-omniverse/issues/432.

Unfortunately changing this attribute triggers a tileset reload because Omniverse doesn't have vertex shader-like functionality. Once Kit 106 is out we should be able to take advantage of point cloud rendering in Fabric.


lilleyse:
image

PR #721, opened 2024-06-25
Update to kit-kernel@106

jshrake-nvidia:
- Updates various build scripts and configuration to compile against kit-kernel@106.
- Replaces individual fabric translation, scale, and orientation values with `omni:fabric:localMatrix` and `omni:fabric:worldMatrix`.
- Fixes a crash in the FabricMaterial destructor.


constant5:
We need this update for our project using IsaacSim 4.5.0.  We have a digital twin we want to locate on earth.   If you need help getting this update working lpease reach at and our team can dedicate some time to it.

lilleyse:
I've been working on this branch on and off and it's in a pretty good state now. Previously with Kit 106.1 some of the MDL samples would consistently crash, but this seems to be fixed in Kit 106.5. I also had to fix how primvar interpolation in specified (i.e. there are no longer `primvars` and `primvarInterpolations` arrays in Kit 106.5).

So all good... except now I get intermittent crashes that I wasn't seeing in Kit 106.1. I'm going to try forcing all prim and texture creation to happen on the main thread to help triage this. Here are some stacktraces:

image (2)

image (1)

image

lilleyse:
> So all good... except now I get intermittent crashes that I wasn't seeing in Kit 106.1. I'm going to try forcing all prim and texture creation to happen on the main thread to help triage this. Here are some stacktraces:

Forcing texture creation to happen on the main thread fixed the crashes. It's not ideal, but it doesn't hurt performance too much.

Issue #727, opened 2024-09-04
Remove _DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR 


corybarr:
Visual Studio 17.10 created some issues with mutexes and memory access in Windows builds. Details are here:https://developercommunity.visualstudio.com/t/Visual-Studio-17100-Update-leads-to-Pr/10669759?sort=newest

Ultimately, we shouldn't need to use the _DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR escape hatch.

lilleyse:
This hack has also been added to the branch of cesium-native that cesium-omniverse uses: https://github.com/CesiumGS/cesium-native/compare/main...cesium-omniverse-build?expand=1

kring:
IMO you should do one of these:

1. Compile Cesium for Omniverse using the same version of Visual Studio that NVidia uses to compile Omniverse itself. This is pretty easy to set up on Github Actions (if you know what version to use), so let me know if you need help. That way, the redistributable installed with Omniverse will work for Cesium, too. This is what we do in Cesium for Unreal.
2. Use the statically-linked version of the MSVCRT, so that the redistributable isn't need at all. This is what we do in Cesium for Unity.
3. Install the right MSVCRT redistributable as part of the the Cesium for Omniverse installation. Not a real strong reason to go with this one that I can think of, but it's an option if you want to use a particular compiler version and don't want to use the static CRT.

Having mismatched CRT versions is likely to cause problems even beyond the very obvious one introduced in 17.10, so I don't think it's a good idea to work around it with `_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR`.

PR #729, opened 2024-09-13
Updated to vcpkg version of cesium-native

lilleyse:
Updates to the latest version of cesium-native, which now uses vcpkg.

Currently targeting the [cesium-omniverse](https://github.com/CesiumGS/cesium-native/commits/cesium-omniverse/) branch because of https://github.com/CesiumGS/cesium-omniverse/issues/727.

The ktx and asyncplusplus ports are copied directly from cesium-unity. The ktx port was especially important for fixing the release vs. debug VC runtime discrepancies we were seeing earlier.

Eventually would like to 

* Fix https://github.com/CesiumGS/cesium-omniverse/issues/727 properly
* Switch to vcpkg for omniverse dependencies
* Use add_subdirectory instead of ExternalProject_add, since that added some complexity here


extern/CMakeLists.txt
```diff
@@ -1,5 +1,231 @@
 include(Macros)
 
+if(MSVC)
```
corybarr, line 3: What if we set `CESIUM_NATIVE_DEPENDENCIES` here to include the common dependencies, then append the platform-specific ones? If we also do this in `src/core/CMakeLists.txt`, that cuts this list down from four occurrences to two.

vcpkg/ports/ktx/portfile.cmake
```diff
@@ -0,0 +1,85 @@
...
+        0004-quirks.patch
+        0005-no-vendored-libs.patch
+        CESIUM-0001-do-not-define-_DEBUG.patch
+        CESIUM-0002-not-a-framework.patch
+        CESIUM-0003-ios-normal-header-path.patch
+)
+file(REMOVE "${SOURCE_PATH}/other_include/zstd_errors.h")
+
+vcpkg_list(SET OPTIONS)
+# if(VCPKG_TARGET_IS_WINDOWS)
```
corybarr, line 20: Do we need to keep this?

corybarr:
> * Use add_subdirectory instead of ExternalProject_add, since that added some complexity here

Is it worth creating an issue for this to improve how handoffable the build system is? I got stuck here, but also haven't managed complex CMake build systems professionally.

extern/CMakeLists.txt
lilleyse: I'll see if we can do that, with the caveat that dependency order sometimes matters.

vcpkg/ports/ktx/portfile.cmake
lilleyse: Probably not... I just copied it straight from Cesium for Unity

lilleyse:
Opened these follow up issues:

* https://github.com/CesiumGS/cesium-omniverse/issues/730
* https://github.com/CesiumGS/cesium-omniverse/issues/731

vcpkg/ports/ktx/portfile.cmake
lilleyse: Actually I think I'll keep the commented block, since it makes it clear how our portfile deviates from https://github.com/microsoft/vcpkg/blob/master/ports/ktx/portfile.cmake

extern/CMakeLists.txt
lilleyse: Fixed in https://github.com/CesiumGS/cesium-omniverse/pull/729/commits/1aaf653a10e77fe572ff4f948facfc57ab024051 with a slightly different approach

vcpkg/ports/ktx/portfile.cmake
corybarr: Sounds good since there's a reason to keep it.

corybarr (approved):
Builds on my system. Code changes look good. Good to go.

Issue #722, opened 2024-08-06
Add support for custom ellipsoids


r-veenstra:
Update Cesium Native and implement support for custom ellipsoids in a method similar to [Cesium for Unreal](https://github.com/CesiumGS/cesium-unreal/pull/1432)

Considerations to be made:

- User needs a way to define and store a custom ellipsoid in their stage
- Existing stages will need to default to a standard WGS84 ellipsoid

PR #732, opened 2024-09-24
Pass custom ellipsoid to tileset and raster overlays

lilleyse:
Follow up to #722.

Passes the custom ellipsoid to `TilesetOptions` and `RasterOverlayOptions`. This fixes tilesets with `region` bounding volumes that use custom ellipsoids. I don't think this has any affect on Cesium Moon Terrain since it uses `box`, but it's a good fix nonetheless.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Change Log
 
+### v0.23.0 - 2024-10-01
+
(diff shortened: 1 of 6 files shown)


corybarr:
Looks good to me.

corybarr (approved):
Looks good to me

PR #733, opened 2024-09-30
Reload tileset if georeference binding changes

lilleyse:
One thing I missed in https://github.com/CesiumGS/cesium-omniverse/pull/732 is to reload the tileset if its georeference binding changes (which implies the ellipsoid may have changed, which requires a tileset reload)

No need to update `CHANGES.md`, falls under the existing mention of ellipsoid fixes.


corybarr (approved):
LGTM

PR #734, opened 2024-10-01
More custom ellipsoid fixes

lilleyse:
Thanks to the `CESIUM_DISABLE_DEFAULT_ELLIPSOID` define, as well looking at some of the native options objects more closely, I was able to fix the rest of the custom ellipsoid issues and moon tiles are no longer being culled.

image


corybarr (approved):
LGTM

PR #735, opened 2024-10-01
Fixed rendering base color textures that contain an alpha channel

lilleyse:
Removed the underlying white color in cesium.mdl.

Before | After
--|--
before|after


corybarr (approved):
LGTM

PR #736, opened 2024-10-01
Fix CI

lilleyse:
CI was broken with the most recent update to `ubuntu-latest`.

* Removed step that was uninstalling system libs. Some of these don't exist anymore and was causing the build to fail. Turns out we don't actually need this step anymore and apparently have enough space.
* Removed CUDA. I'm not sure if this was breaking the build, but we don't need CUDA unless we revisit https://github.com/CesiumGS/cesium-omniverse/pull/459
* Switched to GCC 13. GCC 11 is no longer available on `ubuntu-latest`
* The 6.0.0 version of flake8 was causing issues suddenly. Switched to 7.1.1 and all is good now.


corybarr (approved):
LGTM

lilleyse:
Can't seem to access https://ftpmirror.gnu.org/gnu/m4/m4-1.4.19.tar.xz right now. Will try restarting the build tomorrow.

lilleyse:
Still waiting for the [network issue](https://hostux.social/@fsfstatus/113230041959498522) to resolve, but I'm going to merge this to start the release process.

PR #739, opened 2024-10-04
Fix smooth normals crash

lilleyse:
The latest `moon.usda` from https://github.com/CesiumGS/cesium-omniverse-samples/pull/18 was crashing on windows because of out-of-bounds reads.

Ended up being a pretty clear cut bug in `NormalsAccessor::GenerateSmooth`.


corybarr (approved):
LGTM

Issue #757, opened 2025-10-23
Japan 3D Buildings Data not loading correctly


lilleyse:
The Japan 3D Buildings Data in the Cesium ion asset depot isn't loading correct in Omniverse:

![Image](https://github.com/user-attachments/assets/d9034fe8-95c2-4ea7-824d-27373484bcdf)

Some observations:
* Texture coordinates are within [0, 1] range
* Uses KT2 textures
* Other textured CityGML assets (e.g. Berlin) work fine

Other things to try:

* Try updating cesium-native
* Try tiling other CityGML assets

[japan-buildings.usda.zip](https://github.com/user-attachments/files/23099996/japan-buildings.usda.zip)

Mizuki0522:
@lilleyse @corybarr @CesiumBen 
Just following up if there are any updates on this issue.

PR #758, opened 2025-12-19
Fix loading KTX2 textures with mipmaps

lilleyse:
Fixes https://github.com/CesiumGS/cesium-omniverse/issues/757

The texture loading code was not handling textures with mipmaps. Fixed now.

Before|After
--|--
image|image

#

As a side note, I tried to use `setMipMappedBytesData` but could not get it to work. So for now just the first mip level is being used.

        const auto data = reinterpret_cast<const uint8_t*>(image.pixelData.data());
        const auto dimensions = carb::Uint2{static_cast<uint32_t>(image.width), static_cast<uint32_t>(image.height)};

        if (!image.mipPositions.empty()) {
            std::vector<const uint8_t*> mipMapBytes;
            std::vector<size_t> mipMapStrides;

            mipMapBytes.reserve(image.mipPositions.size());
            mipMapStrides.reserve(image.mipPositions.size());

            for (uint64_t level = 0; level < image.mipPositions.size(); level++) {
                auto levelWidth = glm::max(1, image.width >> level);
                mipMapBytes.push_back(data + image.mipPositions[level].byteOffset);
                mipMapStrides.push_back(4ULL * static_cast<uint64_t>(levelWidth));
            }
            _pTexture->setMipMappedBytesData(
                mipMapBytes.data(), mipMapStrides.data(), image.mipPositions.size(), dimensions, imageFormat);
        } else {
            // As of Kit 105.1, omni::ui::kAutoCalculateStride doesn't work for compressed textures. This value somehow works.
            const auto stride = 4ULL * static_cast<uint64_t>(image.width);
            _pTexture->setBytesData(data, dimensions, stride, imageFormat);
        }

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -138,7 +138,9 @@ jobs:
         if: ${{ runner.os == 'Linux' && matrix.config.name != 'AlmaLinux 8 - GCC' }}
         run: |
           sudo apt update
-          sudo apt install -y doxygen clang-tidy-15 gcovr
+          sudo apt install -y doxygen clang-tidy-15 clang-format-15 gcovr
+          sudo update-alternatives --install /usr/bin/clang-format clang-format /usr/bin/clang-format-15 100
+          sudo update-alternatives --set clang-format /usr/bin/clang-format-15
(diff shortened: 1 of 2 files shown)


Mizuki0522:
Thank you, Sean!

Issue #777, opened 2026-06-18
Cartographic polygon clipping renders tiles black instead of discarding (glTF mask cutout not honored)


harrytruong0804:
## Summary

A `CesiumPolygonRasterOverlay` set to clip (`cesium:overlayRenderMethod = "clip"`) renders the tile geometry **inside** the polygon as **opaque black** instead of discarding it. It appears immediately, and the black geometry keeps depth (it occludes and z-fights / flickers as the camera moves).

## Environment

- Cesium for Omniverse **v0.28.0** (latest release)
- NVIDIA **Isaac Sim 5.1**
- Windows, RTX (Real-Time)

## Steps to reproduce

1. Add a tileset (e.g. Google Photorealistic 3D Tiles).
2. Create a `CesiumCartographicPolygon`.
3. On the tileset, add a `CesiumPolygonRasterOverlay`, bind the polygon, set **Overlay Render Method = clip**.
4. Look at the area inside the polygon.

## Expected

Tiles inside the polygon are discarded → a clean transparent hole (what's behind shows through).

## Actual

Tile geometry inside the polygon renders **opaque black**, and it occludes / z-fights — i.e. the fragments keep depth and are shaded as opaque rather than discarded.

<img width="1920" height="1032" alt="Image" src="https://github.com/user-attachments/assets/4b3f19a0-ee31-478a-9942-8b742a1565c2" />

<img width="1920" height="1032" alt="Image" src="https://github.com/user-attachments/assets/aab6a5fb-4d8e-4af9-b810-481b0d1c1dd5" />

## Root cause

Clipping is implemented as an alpha cutout: `compute_base_color()` returns `float4(0.0)` for clipped texels, and `FabricMaterial::getInitialAlphaMode()` forces the material to glTF `MASK` mode, expecting the renderer to alpha-test-discard (`alpha 0 < alpha_cutoff 0.5`). The discard is delegated to the host's `::gltf::pbr::gltf_material` (`using ::gltf::pbr import *;` in `cesium.mdl`).

When the host's `gltf_material` doesn't translate `alpha_mode = mask` into a working `geometry.cutout_opacity` under RTX, the transparent-black fragment `(0,0,0,0)` is shaded as **opaque**, painting the tile black.

Confirmed during triage:
- `overlayRenderMethod = "clip"` is set (not the `"overlay"` default).
[...]

PR #776, opened 2026-06-11
Update to Kit 110

lilleyse:
Fixes #770 

Updates to the latest version of Kit 110 being used in [kit-app-template](https://github.com/NVIDIA-Omniverse/kit-app-template) and [IsaacSim 6.0.0 dev-2](https://docs.isaacsim.omniverse.nvidia.com/6.0.0/overview/release_notes.html)

No major changes required on our end. Thanks for getting this started @NoeDeMontard!

To do:

* [x] Fix Windows CI
* [x] Update `omni.ui` and `omni.kit.renderer.imgui` hashes on windows

PR #778, opened 2026-06-18
Fix Windows build for Kit 110 (#776)

harrytruong0804:
Addresses the **"Fix Windows CI"** and **"Update `omni.ui` and `omni.kit.renderer.imgui` hashes on windows"** TODO items in #776. Built and verified on Windows against kit-kernel 110.1.1 (NVIDIA Isaac Sim 6.0.0); all three extensions compile, install, and start.

### Changes

- **`extern/nvidia/deps/kit-sdk.packman.xml`** — update the **Windows** `omni.ui` / `omni.kit.renderer.imgui` packman hashes to the Kit 110.1.1 versions:
  - `omni.ui` → `19feb74a05aa1a38` (2.x → **3.1.3**)
  - `omni.kit.renderer.imgui` → `704f5472fcd3d688` (**2.0.6**)
  
  The Windows hashes were still the Kit 109 builds (`omni.ui` 2.29.1), so the C++ plugin was compiled/linked against the wrong `omni.ui` ABI. (Linux hashes were already updated in #776.)

- **`extern/CMakeLists.txt`** — on Windows, Kit 110 ships oneTBB as `tbb12.lib` / `tbb12.dll` (`tbb.dll` no longer exists). Use the versioned name for the TBB prebuilt lib/DLL while keeping the CMake target named `tbb`.

- **`cmake/Macros.cmake`** — expose each prebuilt lib directory via `INTERFACE_LINK_DIRECTORIES` so oneTBB's MSVC auto-link pragma (`#pragma comment(lib, "tbb12.lib")`) resolves at link time (previously `LNK1104: cannot open file 'tbb12.lib'`).

- **`vcpkg/triplets/x64-windows-omniverse.cmake`** — set `VCPKG_CMAKE_CONFIGURE_OPTIONS = -DCMAKE_POLICY_VERSION_MINIMUM=3.5` so vcpkg ports that still declare `cmake_minimum_required(<3.5)` (e.g. `asyncplusplus`) configure under CMake 4.x. *Forward-compat only — not needed on CMake 3.x; drop this hunk if you prefer to pin CMake 3.x.*

### Notes
- The first three are required for any Windows Kit 110 build; the triplet change is only relevant when building with CMake 4.x.
- A separate rendering bug found during testing (clip polygons render black on Kit 110) is tracked in #777 with its own fix PR.


harrytruong0804:
Superseded by https://github.com/CesiumGS/cesium-omniverse/pull/780 — that PR targets `main` directly with the full Kit 110.1.1 update (the #776 changeset plus these Windows build fixes), so it no longer depends on the `kit-110` branch.

